PR #2, opened 2017-08-22
DBZ-127 Tutorial Docker Compose

jpechane:


gunnarmorling:
Works like a charm; thanks a lot, @jpechane!

I'd have two follow-up questions:

* How should we add this to the published tutorial page? I can see how it'd make most sense in terms of logical order to have it at the end. OTOH, we also could add it as some kind of "TL;DR" at the very beginning which would be useful when coming back to the tutorial a second time after having it read once (which I often do myself). I'd lean towards the latter, i.e. having it at the beginning. WDYT?
* Can we add variants of that compose file and instructions for the other connectors, too? In terms of representing it on the website, I e.g. could envision some kind of tabs labelled with "MySQL", "MongoDB", "Postgres", a bit similar to the Maven/Gradle selector on the [Spring website](http://projects.spring.io/spring-data/#quick-start).

jpechane:
* I'd like to combine both approaches. The docker-compose example should be placed at the end. And at the beginning I'd place a notice with the link to the end of the document.
* It is not a problem to do it for all connectors - we just need to seed the initial data so the user can see something. But can we achieve the effect you have described in GitHub readme or would you prefer to have a dedicated page in project website?

gunnarmorling:
> I'd like to combine both approaches. The docker-compose example should be placed at the end. And at the beginning I'd place a notice with the link to the end of the document.

Ok, works for me. Ideally, we could also have a Table of Contents (with links to all headings) at the top of that page. Can be handled separately, though.

> But can we achieve the effect you have described in GitHub readme or would you prefer to have a dedicated page in project website?

I think all the information should be in the tutorial page on the website eventually. But we can start by just adapting the compose files and the example project in the first step.

PR #6, opened 2017-10-04
DBZ-364 OpenShift template usable with barnabas

jpechane:
Do not merge yet - Connect image is not in enmasse Docker repo so we are using a private one


gunnarmorling:
There's an EnMasse image for Connect now, see https://github.com/EnMasseProject/barnabas/pull/12/files.

openshift/README.md
```diff
@@ -0,0 +1,116 @@
...
+It consists of entreprise grade of configuration files and images that brings Kafka on OpenShift.
+
+First we install Kafka broker and Kafka Connect templates into our OpenShift project
+```
+oc create -f https://raw.githubusercontent.com/EnMasseProject/barnabas/master/kafka-statefulsets/resources/openshift-template.yaml
+oc create -f https://raw.githubusercontent.com/EnMasseProject/barnabas/master/kafka-connect/resources/openshift-template.yaml
+oc create -f https://raw.githubusercontent.com/EnMasseProject/barnabas/master/kafka-connect/s2i/resources/openshift-template.yaml
+```
+
+Next we will create a Kafka Connect image with deployed Debezium connectors and deploy a Kafka broker cluster and Kafka Connect cluster
```
gunnarmorling: Instead of building a new image I was hoping the connectors could be added to the existing one externally, e.g. via some mounted volume. What do you think?

openshift/README.md
```diff
@@ -0,0 +1,116 @@
...
+
+Pre-requisities
+ * [Installed and running](https://docs.openshift.org/latest/minishift/getting-started/installing.html) Minishift
+ * [Installed](https://docs.openshift.org/latest/minishift/command-ref/minishift_oc-env.html) OpenShift CLI
+
+
+## Debezium Deployment
+Albeit Debezium comes with its own set of images we are going to re-use Kafka broker and Kafka Connect images that are built and delivered as a part of project [EnMasse](https://github.com/EnMasseProject/).
+One of the componnets of the project is a [Kafka as a Service](https://github.com/EnMasseProject/barnabas/).
+It consists of entreprise grade of configuration files and images that brings Kafka on OpenShift.
```
gunnarmorling: s/entreprise/enterprise/

openshift/README.md
```diff
@@ -0,0 +1,116 @@
...
+    --property print.key=true \
+    --topic dbserver1.inventory.customers
+```
+
+Modify records in the database
+```
+oc exec -it $(oc get pods -o custom-columns=NAME:.metadata.name --no-headers -l app=mysql) -- bash -c 'mysql -u $MYSQL_USER -p$MYSQL_PASSWORD inventory'
+```
+
+## Using Minishift add-ons
```
gunnarmorling: Very nice with those add-ons!

openshift/README.md
jpechane: I'd disagree to this because in this case we would throw away all advantages provided by OpensShift. The barnabas team has add s2i support as a result of discussion with me.
The volumes should be used for data not for binaries as as it is oreferred to have immutable images. Just imagine you have a shared volume. Later on you modify the content but do not restart connect cluster yet. If you then increase number of replicas your cluster will contain replicas with different binaries.
In case of s2i approach - a security fix is released for Debezium.
1. Administrator triggers the binary build with new binaries
2. Openshift builds the new image and upodates the image stream
3. Connect deployment (in a sense of k8s object) automatically detects the change in imagestream and triggers either cluster rolling update or full recereate depending on confiugration
4. A patch is automatically placed in production

The volume apporch is available by barnabas too but it is intended to be used by k8s only as it lacks (currently) the s2i feature.

jpechane:
The exmaple by default deploys cluster of 3 Kafka brokers. If https://github.com/EnMasseProject/barnabas/issues/64 is completed we can reduce it to one broker.

gunnarmorling:
Rebased and applied. Excellent work, @jpechane!

We definitely should blog about it, let's discuss on Gitter where/how/who should do it. debezium.io would be the natural choice, but we also should consider putting out something on developers.redhat.com.

PR #7, opened 2017-10-09
Fix name of the Postgres image

andrewrjones:
The Postgres image specified does not exist. I found https://hub.docker.com/r/debezium/postgres/ and using that works fine, so I assume it's a typo.

That image isn't tagged with a version (other than the Postgres version).

### DIFF
--- a/tutorial/docker-compose-postgres.yaml
+++ b/tutorial/docker-compose-postgres.yaml
@@ -15,7 +15,7 @@ services:
     environment:
      - ZOOKEEPER_CONNECT=zookeeper:2181
   postgres:
-    image: debezium/example-postgres:${DEBEZIUM_VERSION}
+    image: debezium/postgres
     ports:
      - 5432:5432
     environment:


elsmorian:
Heh, I was also going to make this PR, having found that `debezium/example-postgres:0.6` does not exist!

However on using this container, there are no tables in the DB, and the inventory schema does not exist - the tutorial mentions using `--search_path=inventory` as an argument to psql, so I assume the inventory schema is supposed to exist but it doesn't seem to:

```
postgres=# select nspname from pg_catalog.pg_namespace;
      nspname
--------------------
 pg_toast
 pg_temp_1
 pg_toast_temp_1
 pg_catalog
 public
 information_schema
(6 rows)
```

jpechane:
Hi, the problem is here - https://issues.jboss.org/browse/DBZ-383

andrewrjones:
@elsmorian I think that's because those docs are referring to MySQL, which does have the `inventory` schema. Looks like MySQL was the main focus for the tutorial, and Postgres just copy and pasted from there.

Maybe the better fix would be to have a `debezium/example-postgres` image generated that has the same structure as the MySQL one.

elsmorian:
Aha. Yeah A co-worker has just got this example working by making a table name as the whitelist in register-postgres.json and then making changes to it, and seeing the messages come over Kafka, so looks like it just needs a table called 'inventory' to be created to start working.

jpechane:
Till the issue is fixed you can buld the image in question manually - from [here](https://github.com/debezium/docker-images/tree/master/examples/postgres/0.6)

gunnarmorling:
Hi, sorry for the confusion about this one. The tutorial in its original form only targets MySQL and we thought it'd be good to have flavors of it for the other supported DBs, too (see [DBZ-361](https://issues.jboss.org/browse/DBZ-361)). The tutorial uses an example database populated with the `inventory` tables and data. An example image for Postgres has been created already (akin to the one for MySQL), but it's not available yet on DockerHub. We'll add it asap.

gunnarmorling:
Alright, the Postgres example image is available on DockerHub now: https://hub.docker.com/r/debezium/example-postgres/

With that, you can follow the tutorial instructions using Postgres. I'm going to close this PR now, as the image name is correct and the image can be fetched now. Thanks!

jpechane:
@fuxinhao-crypto Hi, I've just tried
```docker pull quay.io/debezium/zookeeper:1.9
1.9: Pulling from debezium/zookeeper
```
And it works as it should

PR #10, opened 2017-11-02
DBZ-430 Adding Compose file and instructions for using Avro messages …

gunnarmorling:
…and the service registry

https://issues.jboss.org/browse/DBZ-430

@jpechane, I was also playing with that UI we were talking about: https://github.com/debezium/debezium-examples/compare/master...gunnarmorling:DBZ-430-ui?expand=1. I don't think it should be part of our examples repo proper (the license is a bit uncommon), but it's nice to see it working with our example.

### DIFF
--- a/tutorial/README.md
+++ b/tutorial/README.md
@@ -80,3 +80,38 @@ db.customers.insert([
 # Shut down the cluster
 docker-compose -f docker-compose-mongodb.yaml down
 ```
+
+## Using MySQL and the Avro message format
+
+To use [Avro-style messages](http://debezium.io/docs/configuration/avro/) instead of JSON,
+follow the instructions for MySQL above,
+but use the _docker-compose-mysql-avro.yaml_ configuration file instead.
+This Compose file configures the Connect service to use the Avro (de-)serializers and starts one more service,
+the Confluent service registry.
+Using Avro on conjunction with the service registry allows for much more compact messages.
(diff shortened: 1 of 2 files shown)


tutorial/README.md
```diff
@@ -80,3 +80,32 @@ db.customers.insert([
...
+This Compose file configures the Connect service to use the Avro (de-)serializers and starts one more service,
+the Confluent service registry.
+Using Avro on conjunction with the service registry allows for much more compact messages.
+
+You can access the first version of the schema for `customers` values like so:
+
+```shell
+curl -X GET http://localhost:8081/subjects/dbserver1.inventory.customers-value/versions/1
+```
+
```
jpechane: It might be worthy to add this command too
```
curl -X GET http://localhost:8081/subjects/dbserver1.inventory.customers-value/versions/1 | jq '.schema | fromjson'
```
This prints the schema in pretty-print.

jpechane:
Nice work, thanks! No changes needed, jsut check my proposal for a command to print schema.

tutorial/README.md
gunnarmorling: Good idea, pushed an update. I normally use httpie instead of curl, but jq can handle the excaped JSON nicely here.

jpechane:
Merged, thanks.

PR #15, opened 2018-01-20
Fix connect-jdbc-es folder rename

rmoff:
Current docker-compose files are not working because the `connect-jdbc` folder has been renamed to `connect-jdbc-es`


gunnarmorling:
Merged. Good catch, thanks a lot, @rmoff!

PR #16, opened 2018-01-23
Use Strimzi 0.1.0 release instead of master / latest

scholzj:
The examples should not be using files from the Strimzi master branch, which might be changing. It should use the release instead. This PR changes the links to the files from master to the 0.1.0 release. It also changes the tag for Docker images from `latest` to 0.1.0 and fixes the link to Strimzi website.


gunnarmorling:
Applied. Thanks, @scholzj!

PR #20, opened 2018-02-27
add first basic kstreams version as described in the upcoming blog post

hpgrahsl:
@gunnarmorling as discussed here is the simple version referenced from the 1st upcoming blog post

### DIFF
--- a/kstreams/docker-compose.yaml
+++ b/kstreams/docker-compose.yaml
@@ -1,0 +1,67 @@
+version: '2'
+services:
+  zookeeper:
+    image: debezium/zookeeper:${DEBEZIUM_VERSION}
+    ports:
+     - 2181:2181
+     - 2888:2888
+     - 3888:3888
+  kafka:
+    image: debezium/kafka:${DEBEZIUM_VERSION}
+    ports:
+     - 9092:9092
+    links:
+     - zookeeper
+    environment:
+     - ZOOKEEPER_CONNECT=zookeeper:2181
+  mysql:
+    image: debezium/example-mysql:${DEBEZIUM_VERSION}
+    ports:
+     - 3306:3306
+    environment:
+     - MYSQL_ROOT_PASSWORD=debezium
+     - MYSQL_USER=mysqluser
+     - MYSQL_PASSWORD=mysqlpw
+  mongodb:
+    image: mongo:latest
+    ports:
+     - 27017:27017
+  connect_source:
+    image: debezium/connect:${DEBEZIUM_VERSION}
+    build:
+      context: debezium-mongodb
+    ports:
+     - 8083:8083
+    links:
+     - kafka
+     - mysql
+    environment:
+     - BOOTSTRAP_SERVERS=kafka:9092
+     - GROUP_ID=1
+     - CONFIG_STORAGE_TOPIC=my_source_connect_configs
+     - OFFSET_STORAGE_TOPIC=my_source_connect_offsets
+  connect_sink:
+    image: debezium/connect-mongodb:${DEBEZIUM_VERSION}
+    build:
+      context: debezium-mongodb
+    ports:
+     - 8084:8083
+    links:
+     - kafka
+     - mongodb
+    environment:
+     - BOOTSTRAP_SERVERS=kafka:9092
+     - GROUP_ID=2
+     - CONFIG_STORAGE_TOPIC=my_sink_connect_configs
+     - OFFSET_STORAGE_TOPIC=my_sink_connect_offsets
+     - CONNECT_KEY_CONVERTER_SCHEMAS_ENABLE=false
+     - CONNECT_VALUE_CONVERTER_SCHEMAS_ENABLE=false
+  aggregator:
+    image: debezium/kstreams-aggregator:${DEBEZIUM_VERSION}
+    build:
+      context: poc-ddd-aggregates
+    links:
+     - kafka
+    environment:
+     - JAVA_MAIN_CLASS=com.github.hpgrahsl.kafka.StreamingAggregatesDDD
+    command: /deployments/run-java.sh dbserver1.inventory.customers dbserver1.inventory.addresses kafka:9092
--- a/kstreams/docker-compose.yaml
+++ b/kstreams/docker-compose.yaml
@@ -1,4 +1,4 @@
-version: '2'
+version: '2.1'
 services:
   zookeeper:
     image: debezium/zookeeper:${DEBEZIUM_VERSION}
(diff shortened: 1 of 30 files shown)


kstreams/poc-ddd-aggregates/pom.xml
```diff
@@ -0,0 +1,63 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <groupId>com.github.hpgrahsl.kafka</groupId>
```
gunnarmorling: Could we move it under `io.debezium.examples`?

kstreams/poc-ddd-aggregates/src/main/java/com/github/hpgrahsl/kafka/StreamingAggregatesDDD.java
```diff
@@ -0,0 +1,100 @@
+package com.github.hpgrahsl.kafka;
```
gunnarmorling: Could we rename it into `io.debezium.examples.aggregation`?

kstreams/poc-ddd-aggregates/src/main/java/com/github/hpgrahsl/kafka/serdes/JsonHybridDeserializer.java
```diff
@@ -0,0 +1,58 @@
...
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.github.hpgrahsl.kafka.model.EventType;
+import org.apache.kafka.common.errors.SerializationException;
+import org.apache.kafka.common.serialization.Deserializer;
+
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+
+public class JsonHybridDeserializer<T> implements Deserializer<T> {
```
gunnarmorling: I'm trying to figure out the purpose of this deserializer; in which case could something be DBZ JSON _or_ one of the mapped classes? Perhaps two different deserializers would be better, one for handling DBZ's JSON and one for the other case?

Btw. it might be useful to use the existing [flatting SMT](http://debezium.io/docs/configuration/event-flattening/) for extracting just the after state of a CDC event. WDYT?

kstreams/poc-ddd-aggregates/src/main/java/com/github/hpgrahsl/kafka/StreamingAggregatesDDD.java
```diff
@@ -0,0 +1,100 @@
...
+                            latest.update(address,addressId,new DefaultId(address.getCustomer_id()));
+                            return latest;
+                        },
+                        Materialized.as(childrenTopic+"_table_temp")
+                                .withKeySerde((Serde)defaultIdSerde)
+                                    .withValueSerde(latestAddressSerde)
+                );
+
+        //2b) aggregate addresses per customer id
+        KTable<DefaultId, Addresses> addressTable = tempTable.toStream()
```
gunnarmorling: There are some warnings about type safety when looking at this in Eclipse. Any idea how those could be avoided?

kstreams/poc-ddd-aggregates/pom.xml
hpgrahsl: sure. will do

kstreams/poc-ddd-aggregates/src/main/java/com/github/hpgrahsl/kafka/StreamingAggregatesDDD.java
hpgrahsl: dito

kstreams/poc-ddd-aggregates/src/main/java/com/github/hpgrahsl/kafka/serdes/JsonHybridDeserializer.java
hpgrahsl: while you are right that it can be done with separate Serdes I decided for combining this for reasons of brevity. otherwise you would need to instantiate and specify more serdes for the kstreams topology. would recommend to leave it like that for the PoC / blog post sample.

kstreams/poc-ddd-aggregates/src/main/java/com/github/hpgrahsl/kafka/StreamingAggregatesDDD.java
hpgrahsl: good point, I suspect something related to lambda type inference limitations? most likely explicit typing should help. I'll look into this!

hpgrahsl:
PR updated to clean up your suggestions!

kstreams/poc-ddd-aggregates/src/main/java/com/github/hpgrahsl/kafka/StreamingAggregatesDDD.java
hpgrahsl: this should also be fixed now

gunnarmorling:
Thanks for the update, @hpgrahsl! I've pushed one commit to this PR for the Docker Compose set-up. 

There's one problem with it, though: the aggregator process fails because that Java program is launched before the source topics have been created by the Debezium connector. Do you know whether there's a way so that KStreams won't fail but will just wait if a required topic doesn't exist yet?

Note that this also requires your changes to the examples SQL file in the 0.7 MySQL example image. I'll apply them upstream, but if you wanted to try this Compose Set-Up yourself, you'd have to update that image locally yourself for the time being.

hpgrahsl:
Awesome that you have made all the docker additions!

W.r.t the kstreams app failing I'll try to take a look ... hopefully tomorrow. If there is no other option we could check for topic existance and if not available pre-create them explicitly from within the kstreams app using the kafka admin api. however, in general it is recommended that all topics are created before running any kafka apps. even for running kafka connect it's not a good practice to let kafka brokers auto-create topics as needed but instead have them explicitly pre-created beforehand.

anyway, I'll keep you updated.

gunnarmorling:
Found a good enough way and pushed another commit for that: the aggregator service invokes curl now for checking whether the source connector has been deployed; only if that's the case it'll continue and launch the aggregator. I think that's good enough for our purposes.

kstreams/docker-compose.yaml
```diff
@@ -63,5 +63,16 @@ services:
...
-     - JAVA_MAIN_CLASS=com.github.hpgrahsl.kafka.StreamingAggregatesDDD
-    command: /deployments/run-java.sh dbserver1.inventory.customers dbserver1.inventory.addresses kafka:9092
+     - JAVA_MAIN_CLASS=io.debezium.examples.aggregation.StreamingAggregatesDDD
+    depends_on:
+     - connect_source
+    command: >
+      /bin/bash -c "
+        while ! curl -f -s -X GET http://connect_source:8083/connectors/mysql-source;
+        do
+          echo 'Waiting for sink connector to be deployed';
```
hpgrahsl: echo should read 'source connector' instead of sink?

kstreams/docker-compose.yaml
```diff
@@ -63,5 +63,16 @@ services:
...
+          sleep 1;
+        done;
+        echo 'Sink connector deployed!';
```
hpgrahsl: dito for this echo

gunnarmorling:
Agh, yes, of course. Thanks for spotting. Seems it was a bit late last
night :)

gunnarmorling:
Rebased and applied. Thanks a lot, @hpgrahsl!

PR #22, opened 2018-03-08
WIP: First cut of Kinesis example

jpechane:


kinesis/README.md
```diff
@@ -0,0 +1,104 @@
+# Streaming database changes to Amazon Kinesis using Debezium
+
+[Debezium](http://debezium.io/) was a set of Apache Kafka connectors that allows streaming of change events from different database. Using [Debezium Embedded](http://debezium.io/docs/embedded/) it is possible to stream database changes to an arbitrary destination and not be limited to Kafka broker only.
```
gunnarmorling: s/was/is/
s/different database/multiple databases/

kinesis/README.md
```diff
@@ -0,0 +1,104 @@
...
+
+This demo shows how to stream changes from MySQL database running on a local machine to aa Amazon [Kinesis](https://aws.amazon.com/kinesis/data-streams/) stream.
```
gunnarmorling: s/aa/an/

kinesis/README.md
```diff
@@ -0,0 +1,104 @@
...
+
+## Pre-requisities
+* Java 8 development environment
+* Local Dockar installation to run source database
```
gunnarmorling: s/Dockar/Docker/
"the source database"

kinesis/README.md
```diff
@@ -0,0 +1,104 @@
...
+* an Amazon [AWS](https://aws.amazon.com/) account
+* [AWS CLI](https://aws.amazon.com/cli/) client
+
+## How to run
+### Start MySQL source database
+We will start a pre-populated MySQL database that is the same as used by Debezium [tutorial](http://debezium.io/docs/tutorial/).
```
gunnarmorling: "by the"

kinesis/README.md
```diff
@@ -0,0 +1,104 @@
...
+
+```mvn docker:run```
+
+### Initialize Kinesis stream
+We suppose that you have already executed `aws configure` as descirbed in AWS CLI installation guide.
```
gunnarmorling: s/descirbed/described/

A link to the install guide would be nice.

kinesis/README.md
```diff
@@ -0,0 +1,104 @@
...
+
+### Create the Kinesis stream
+
+```aws kinesis create-stream --stream-name debezium --shard-count 1```
+
+We are using only one shard to keep the events from database in ordered sequence. If you are not interested in a total order but the order per table is enough you can modify the code to use the table name as a sharding key and increase the number of shards.
```
gunnarmorling: So Kinesis does guarantee order of messages within a shard?

kinesis/src/main/java/io/debezium/examples/kinesis/ChangeDataSender.java
```diff
@@ -0,0 +1,88 @@
...
+                .withCredentials(credentialsProvider)
+                .withRegion(regionName)
+                .build();
+    }
+
+    public void run() {
+        final EmbeddedEngine engine = EmbeddedEngine.create()
+                .using(config)
+                .using(this.getClass().getClassLoader())
+                .using(this.getClass().getClassLoader())
```
gunnarmorling: double call

kinesis/src/main/java/io/debezium/examples/kinesis/ChangeDataSender.java
```diff
@@ -0,0 +1,88 @@
...
+                .using(Clock.SYSTEM)
+                .notifying(this::sendRecord)
+                .build();
+        engine.run();
+    }
+
+    private void sendRecord(SourceRecord record) {
+        // We are interested only in data events not schema change events
```
gunnarmorling: Why do we see those here? Is it the (internal DB history topic)? Wouldn't it make sense then to use another implementation (the file based one)?

kinesis/README.md
gunnarmorling: Thinking more about that, what would be needed to have a stream per source table, more closely mirroring the behavior you'd get with Kafka by default?

kinesis/src/main/java/io/debezium/examples/kinesis/ChangeDataSender.java
```diff
@@ -0,0 +1,88 @@
...
+
+    public void run() {
+        final EmbeddedEngine engine = EmbeddedEngine.create()
+                .using(config)
+                .using(this.getClass().getClassLoader())
+                .using(this.getClass().getClassLoader())
+                .using(Clock.SYSTEM)
+                .notifying(this::sendRecord)
+                .build();
+        engine.run();
```
gunnarmorling: How is it stopped? I'd suggest to run it within an executor (as suggested in the embedded docs), block the main thread while waiting for a key press on the console and call `engine.stop()` once that has happened.

kinesis/src/main/java/io/debezium/examples/kinesis/ChangeDataSender.java
```diff
@@ -0,0 +1,88 @@
...
+
+    private void sendRecord(SourceRecord record) {
+        // We are interested only in data events not schema change events
+        if (record.valueSchema().name().endsWith(".Envelope")) {
+            final byte[] payload = converter.fromConnectData("dummy", record.valueSchema(), record.value());
+
+            PutRecordRequest putRecord = new PutRecordRequest();
+            putRecord.setStreamName(streamName);
+            putRecord.setPartitionKey(APP_NAME); // We want events to keep total ordering
+            putRecord.setData(ByteBuffer.wrap(payload));
```
gunnarmorling: Shouldn't we propagate the key as well?

kinesis/src/main/java/io/debezium/examples/kinesis/ChangeDataSender.java
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    private void sendRecord(SourceRecord record) {
+        // We are interested only in data events not schema change events
+        if (record.valueSchema().name().endsWith(".Envelope")) {
+            final byte[] payload = converter.fromConnectData("dummy", record.valueSchema(), record.value());
+
+            PutRecordRequest putRecord = new PutRecordRequest();
+            putRecord.setStreamName(streamName);
+            putRecord.setPartitionKey(APP_NAME); // We want events to keep total ordering
```
gunnarmorling: Seems setting the table/topic name would be a more natural partition key for typical Debezium usage?

kinesis/src/main/java/io/debezium/examples/kinesis/ChangeDataSender.java
jpechane: I was thinking more in terms of shutdown hook instred of waiting for key press.

kinesis/src/main/java/io/debezium/examples/kinesis/ChangeDataSender.java
jpechane: Yes, but I wanted to make demo simple and TBH cheaper - one pays per-stream.

kinesis/src/main/java/io/debezium/examples/kinesis/ChangeDataSender.java
jpechane: No, the internal DB history is configured to be in memory. This is schema change event topic.

jpechane:
@gunnarmorling I've tried to incroporate your comments. I've also changed the scenario a little bit. We now stream table per Kinesis stream but to keep things simple I set a filter to only one table

gunnarmorling:
@jpechane let's talk about this once the 0.8 release is out. Would be nice to have it on the blog soon. Thanks!

gunnarmorling:
@jpechane Now that 0.8 Final is out, it's time to get back to this one. I'll spend some time on it next week.

kinesis/src/main/java/io/debezium/examples/kinesis/ChangeDataSender.java
gunnarmorling: Ok, seems easier then to just suppress the topic by setting `include.schema.changes` to `false`. I'll apply some updates to this PR.

gunnarmorling: Sorry btw. for the huge delay of reviewing this :( But I'm going to focus on it finally now.

kinesis/README.md
```diff
@@ -0,0 +1,105 @@
...
+Start the application that uses Debezium Embedded to get change events from database to Kinesis stream.
+```
+mvn exec:java
+```
+
+### Read events from the stream
+Execute command
+
+```
+aws kinesis get-records --shard-iterator $ITERATOR | jq -r '.Records[].Data' | base64 -d | jq .
```
gunnarmorling: I'm having trouble with executing this line. Specifically, I'm wondering how the _base64_ invocation is working for you; the previous _jq_ call yields an array of lines, for each of which _base64_ must be invoked individually. Here it appears though that the entire array is passed at once, causing _base64_ to fail after the first line. Instead it'd need looping through. I'm curious how this works for you; could it be that _base64_ is more capable on Linux that on OS X?

gunnarmorling: Ok, turns out that _jq_ has built-in base64 support in 1.6:

    aws kinesis get-records --shard-iterator $ITERATOR | jq -r '.Records[].Data | @base64d'

gunnarmorling:
Rebased and applied. Thanks a lot, @jpechane! I'm going to make some more updates and will get started with the accompanying blog post. My goal is to publish it tomorrow.

gunnarmorling:
@jpechane I'm having some doubts about using the record id as the Kinesis partition key. Its semantics are different from the message key in Kafka (it's not used for compaction etc.) and most importantly, it's limited in length to 256 bytes.

So I'm inclined to use a hash of the record key as partition key value. Any thoughts?

PR #23, opened 2018-06-16
Updated README.md

koutsoumposval:
* Run docker-compose in background
* Exported CURRENT_HOST variable for docker-machine virtual & local machine
* Amended connector commands in order to use CURRENT_HOST variable


jpechane:
@koutsoumposval Applied, thanks!

PR #24, opened 2018-07-13
Updating versions after 0.8.0.Final release

gunnarmorling:
@jpechane A quick one after the 0.8.0.Final release :)


jpechane:
@gunnarmorling Rebased and applied, thanks!

PR #25, opened 2018-07-23
DBZ-40 SQL Server usage example

jpechane:
https://issues.jboss.org/browse/DBZ-40


gunnarmorling:
Very nice; applied. Thanks, @jpechane!

PR #28, opened 2018-09-12
Exploration of aggregate materialization with JPA

gunnarmorling:
*Update:* @jpechane, the PR is cleaned up much more now, also the README provides some basic instructions. So you could give it a try and let me know what you think :)

Hey @jpechane, here's what I've been up to these past days.

The idea is to materialize aggregates (say a customer and their addresses) within a dedicated table in the source DB and within the upstream TX. DBZ then is used to only CDC this single table, and a new SMT is used to convert the aggregates (which have been persisted as JSON) into properly typed Kafka Connect records which then can be sinked e.g. using the Elasticsearch sink connector or @hpgrahsl's MongoDB connector.

It's still very rough, but the main parts are working. To try it out, do this (the README still is not up to date):

* Build the SMT: mvn clean install -f json-smt-es/pom.xml
* docker-compose up --build
* curl -i -X POST -H "Accept:application/json" -H  "Content-Type:application/json" http://localhost:8083/connectors/ -d @source.json
* Import the _jpa-test_ project into your IDE and run `JpaAggregationTest`; this updates a customer, causing the materialization of an aggregate of the customer and all associated records in other tables
* Examine the aggregate on the console: docker-compose exec kafka /kafka/bin/kafka-console-consumer.sh \
    --bootstrap-server kafka:9092 \
    --from-beginning \
    --property print.key=true \
    --topic customers-complete
* Sink aggregate into ES: curl -i -X POST -H "Accept:application/json" -H  "Content-Type:application/json" http://localhost:8083/connectors/ -d @es-sink-aggregates.json
* curl -X GET -H "Accept:application/json" http://localhost:9200/customers-complete/_search?pretty

On the JPA side, I've built a small Hibernate extension which evaluates the `@MaterializeAggregate` annotation (see `Customer`) and creates a JSON representation of the annotated entity (using Jackson). The most challenging part was to have Jackson emit Kafka Connect Schema instead of regular JSON Schema. That's vital though to not loose type info (e.g. correct numeric types over JSON's very limited native types).

I'll polish and improve going forward, any feedback is more than welcome. Let me know if you're trying to run it and have any issues. As said, still need to clean up and provide more detailed instructions :)

### DIFF
--- a/jpa-aggregations/README.md
+++ b/jpa-aggregations/README.md
@@ -1,0 +1,48 @@
+# Debezium Hibernate Aggregate Materialization Demo
+
+This demo shows how to materialize consistent aggregates (e.g. a customer and all their addresses) using a PoC-level Hibernate ORM extension.
+This extension persists materialized aggregates (represented as JSON) in a dedicated table, `aggregates`.
+Debezium is set up to capture changes from this table and stream them into Kafka.
+An SMT (single message transform) is used to expand the aggregate's JSON into typed Kafka Connect records and route them into a dedicated topic per aggregate root type.
+The Elasticsearch sink connector is used to consume these records and persist the structured aggregate into Elasticsearch.
+
+## Usage
+
+How to run:
+
+```shell
+# Start the DB, Kafka Connect, Elasticsearch etc.
+export DEBEZIUM_VERSION=0.9
+docker-compose up --build
+
+# Register MySQL connector to capture changes from the "aggregates" table
+curl -i -X POST -H "Accept:application/json" -H  "Content-Type:application/json" http://localhost:8083/connectors/ -d @source.json
+```
+
+Import the _jpa-test_ project into your IDE and execute `JpaAggregationTest`.
+
+```shell
+# Observe changes to the aggregate topic while applying more changes to customers using the test class above
+docker-compose exec kafka /kafka/bin/kafka-console-consumer.sh \
+    --bootstrap-server kafka:9092 \
+    --from-beginning \
+    --property print.key=true \
+    --topic customers-complete
+```
+
+```shell
+# Register ES sink connector
+curl -i -X POST -H "Accept:application/json" -H  "Content-Type:application/json" http://localhost:8083/connectors/ -d @es-sink-aggregates.json
+```
+
+Examine contents of the Elasticsearch index while you alter the customer data:
+
+```shell
+curl -i -X GET -H "Accept:application/json" http://localhost:9200/customers-complete/_search?pretty
+```
+End the application:
+
+```shell
+# Shut down the cluster
+docker-compose down
+```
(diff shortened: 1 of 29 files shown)


gunnarmorling:
Btw. I should point out some limitations of the approach in general and the current implementation:

* In general, it's driven by the access layer, i.e. if a user bypasses Hibernate (plain JDBC, direct DB access), naturally this solution doesn't work
* Types might differ when comparing a field from the aggregate and a CDC message directly obtained from Debezium tracking the table; not a big issue in practive I suppose but it should be pointed out

* The implementation assumes there's always a numeric (long) id in involved entities; obviously, this would have to be made more generic, with aggregate tables specific to individual PK formats.
* When only updating a referenced entity (say an address' ZIP code) without touching the root aggregate, we'll miss that. Support for this could be added in general, if we wanted to provide this as an official DBZ component

gunnarmorling:
As I'm thinking more about this idea, I just realized an important limitation: the materialization listener won't be triggered in case of bulk updates/deletes via the Criteria API and JPQL queries. For that, a manual materializing API and/or batch will be needed.

gunnarmorling:
> * The implementation assumes there's always a numeric (long) id in involved entities; obviously, this would have to be made more generic, with aggregate tables specific to individual PK formats.

In fact I think we just could address this by serializing the key into JSON, too.

jpechane:
@gunnarmorling FYI - Developer Studio (and JBoss Tools probably too) JPA validator goes mad with a lot of issues so it should probably be taken into account during the final polishing.

jpechane:
@gunnarmorling Would it be a problem to introduce a dependency on Connect? I mean instead of making your own serializer you could convert Object into `Struct` and then apply either JSON or Avro converter to do the dirty work?

hpgrahsl:
@jpechane while it would definitely simplify a few things it directly introduces coupling to connect api just for building these aggregates. whether this is good/bad depends a bit of where you plan to go in the future with this. personally I would stick to @gunnarmorling approach because the aggreates building is something that should be possible in (any existing) applications without a connect 'bias'.

gunnarmorling:
> FYI - Developer Studio (and JBoss Tools probably too) JPA validator goes mad with a lot of issues so it should probably be taken into account during the final polishing.

Can you share some details of the issues it reports?

gunnarmorling:
> Would it be a problem to introduce a dependency on Connect? I mean instead of making your own serializer you could convert Object into Struct and then apply either JSON or Avro converter to do the dirty work?

How would that look like? I.e. how would the conversion of Object to Struct look like? Unless there's some functionality for that already, that'd be lots of work. Currently, Jackson is doing all the heavy-lifting here. Also all its features (e.g. `@JsonIgnore`) can be used. The custom serializers I created are just there for exporting Connect's JSON schema notion instead of regular [JSon Schema](https://json-schema.org/), actual object serialization is plain Jackson.

gunnarmorling:
Reworked this a bit so that the id of materialized aggregates now also is a serialized JSON representation. I think with that it's good as a PoC and ready to be merged, sans some minor clean-up. Going to work on the blog post now.

gunnarmorling:
Merging this now. Thanks for your feedback, @jpechane and @hpgrahsl!

PR #30, opened 2018-09-26
[kstreams] Demo for merging two CDC streams with KStreams and expose the

gunnarmorling:
... aggregated results to a WebSocket client

Still somewhat work-in-progress, but as @hpgrahsl was so curious, let me share it early. 😎  The README.md has all the instructions. /CC @jpechane.


hpgrahsl:
@gunnarmorling very nice demo indeed! here are a few improvement suggestions:
1) rename variable maxTemperaturesByStation -> avgTemperatureByStation
2) try to use generic wrapper serdes in order to get rid of some boilerplate serde code
3) writing values directly to the websocket in the .peek(...) method is special :)
4) it might make sense to add / change to window-based averages for nicer graphs on the client UI

gunnarmorling:
> writing values directly to the websocket in the .peek(...) method is special :)

@hpgrahsl So what would be the "proper" way to do this? I guess I could write to a topic and then use a WS sink connector (surely one exists), but it adds extra complexity, esp. due cross-domain requests (that connector would run within KC and thus another host). So I was wondering what would be the best to do this?

kstreams-live-update/temperature-aggregator/src/main/java/io/debezium/examples/kstreams/liveupdate/aggregator/Main.java
```diff
@@ -1,102 +1,55 @@
...
-        maxTemperaturesByStation.toStream().to("average_temperatures_by_station", Produced.with(Serdes.String(), Serdes.String()));
-        maxTemperaturesByStation.toStream().print(Printed.toSysOut());
-
-        final KafkaStreams streams = new KafkaStreams(builder.build(), props);
-        Runtime.getRuntime().addShutdownHook(new Thread(streams::close));
-
-        streams.start();
+  public static void main(String[] args) {
+
+    if (args.length != 1) {
```
gunnarmorling: Thanks for cleaning this up btw!

First missed it because all the indentation changed ;)

gunnarmorling:
Rebased and applied. Thanks a lot for your comments and work, @hpgrahsl and @hpgrahsl! I'll leave the change to use a JSON SerDe for another time. I'll update it to use a windowed query before I do the demo for the next time (Kafka Summit).

Btw. in terms of the example domain I'm contemplating now to make it an `Order` and a `Region` table, and then aggregate to sliding average revenue per regions. This might feel a bit less contrived as you'd naturally store orders in a database, whereas for temperature sensors one might argue that those could send their values to Kafka right away.

PR #31, opened 2018-10-01
Avoid NullPointerExceptions if data is null

nilshartmann:
When `data` is null (which is fine according to API doc), `readTree` fails with a NullPointerException.

### DIFF
--- a/kstreams-live-update/temperature-aggregator/src/main/java/io/debezium/examples/kstreams/liveupdate/aggregator/serdes/ChangeEventAwareJsonSerde.java
+++ b/kstreams-live-update/temperature-aggregator/src/main/java/io/debezium/examples/kstreams/liveupdate/aggregator/serdes/ChangeEventAwareJsonSerde.java
@@ -7,6 +7,7 @@ package io.debezium.examples.kstreams.liveupdate.aggregator.serdes;
 
 import java.io.IOException;
 import java.util.Map;
+import java.util.logging.Logger;
 
 import org.apache.kafka.common.serialization.Deserializer;
 import org.apache.kafka.common.serialization.Serde;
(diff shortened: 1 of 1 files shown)


gunnarmorling:
LGTM; out of curiousity, though, in which case though did you see a null as
value? And could you prefix the commit message with "[kstreams] ..."?

nilshartmann:
tbh I cannot reproduce the case. I think I ran docker-compose, deployed the mysql connector, waited for wildfly to be ready and then opened the website with your ws example. Now the null data arrived, but only for some amount of time. After some moments everything worked as expected an the NPE disappeared.

As I think receiving null values in the example might be a bug, I added a log warning so that one sees when the problem occures again.

kstreams-live-update/temperature-aggregator/src/main/java/io/debezium/examples/kstreams/liveupdate/aggregator/serdes/ChangeEventAwareJsonSerde.java
```diff
@@ -29,6 +30,8 @@
  */
 public class ChangeEventAwareJsonSerde<T> implements Serde<T> {
 
+    Logger log = Logger.getLogger( this.getClass().getName() );
```
gunnarmorling: Can you change this into `private static final Logger LOG = ...`?

kstreams-live-update/temperature-aggregator/src/main/java/io/debezium/examples/kstreams/liveupdate/aggregator/serdes/ChangeEventAwareJsonSerde.java
```diff
@@ -67,6 +70,13 @@ public void configure(Map<String, ?> configs, boolean isKey) {
 
         @Override
         public T deserialize(String topic, byte[] data) {
+
```
gunnarmorling: Can you lease remove this new line and indent with four spaces?

gunnarmorling:
Closing this, the change is incorpated in #32, too. Thanks for your efforts, @nilshartmann!

PR #33, opened 2018-10-04
Add GraphQL Example

nilshartmann:
Just a first idea: a GraphQL API that allows clients to subscribe to Debezium Change Events.

Hopefully the README contains all information to get started.

Some thoughts:

* I think the `aggregator` app should be renamed to something more appropriate. If anyone has an idea, please let me know, I'm quite bad in finding names...

* The GraphiQL API browser seems to not show all of the received messages. One can see in the Network inspector that it receives all messages/events from the server but only displays the last one out of a "batch". I've added a simple cli tool that can execute GraphQL subscriptions and show all results on the console. 

* Also I'm not sure If the current GraphQL API should be enhanced so that clients can query also the associated Customer and Product objects for an Order (Maybe via a REST endpoint in _event-source_, which then could be renamed to _shop_ or something like this). Currently clients can only query the IDs of the referenced objects:
```
subscription {
 onNewOder {
  productId
  customerId
  quantity
 }
}
````

"Normally" in a GraphQL API you would be able to fetch whole domain object graphs like so:
```
subscription {
 onNewOder {
  product { id name }
  customer { id email }
  quantity
 }
}
````

(Could lead to some interesting performance problems btw... 😉)


gunnarmorling:
Rebased and applied. Added two more commits for some clean-up. Thanks a ton for working on this, @nilshartmann, it's really fantastic!

Regarding

> clients can query also the associated Customer and Product objects for an Order

That's a very interesting topic, but it'll will be a bit more complex. The reason being that Debezium gives you one topic per table, so you'd have to use KStreams or similar to produce the join of orders and customers in order to be able to let the user select customer properties from the order aggregate via GraphQL. Definitely not needed for the first demo IMHO.

PR #37, opened 2018-10-24
Kstreams improvements

gunnarmorling:
@jpechane These are some improvements to the KStreams demo:

* The existence of topics is now checked programmatically within the aggreator itself; this makes it easier to run the demo on OpenShift, as the shell script for checking the topics isn't needed any longer
* Messages without JSON schema can be handled by the aggregator now; again this makes it nicer to demo this, as messages in the console consumer are more comprehensible.


kstreams-live-update/aggregator/src/main/java/io/debezium/examples/kstreams/liveupdate/aggregator/StreamsPipeline.java
```diff
@@ -67,4 +77,35 @@
                         .divide(BigDecimal.valueOf(100), 2, RoundingMode.HALF_UP))
                 .mapValues(v -> String.valueOf(v));
     }
+
```
jpechane: Do we really need to check the presence of the topics?

gunnarmorling: Yes, I think so. The KStreams pipeline fails otherwise. That's why there was the shell step in it before.

kstreams-live-update/aggregator/src/main/java/io/debezium/examples/kstreams/liveupdate/aggregator/cdi/EagerExtension.java
```diff
@@ -0,0 +1,47 @@
+/*
```
jpechane: I must say I am not happy with all the CDI extension stuff in the demo as it is more or less distraction from the main point of the example. Cannot we use `@Initialized` observer to achive the same functionality?

jpechane:
@gunnarmorling Overall I think it is fine, I've just raised a concer about CDI (over-)use.

kstreams-live-update/aggregator/src/main/java/io/debezium/examples/kstreams/liveupdate/aggregator/cdi/EagerExtension.java
gunnarmorling: I hear what you're saying, but it's the best I could come up with. If you got ideas for improving it, why not. Otherwise I think it's fine as is.

kstreams-live-update/aggregator/src/main/java/io/debezium/examples/kstreams/liveupdate/aggregator/StreamsPipelineManager.java
```diff
@@ -0,0 +1,101 @@
...
+    private String kafkaBootstrapServers;
+
+    @Inject
+    private ChangeEventsWebsocketEndpoint websocketsEndPoint;
+
+    private KafkaStreams streams;
+    private ExecutorService executor;
+
+    @PostConstruct
+    public void startKStreams() {
```
jpechane: Replace 
```
    @PostConstruct
    public void startKStreams() {
```
with `public void startKStreams(@Observes @Initialized(ApplicationScoped.class) Object init) {`

and there is no need for the eager stuff.

gunnarmorling: Hum, so I have tried that, but without luck. The method isn't invoked, which makes sense as the application scoped bean will be loaded lazily. So I'm going to go with the original proposal, happy to adjust if there's an easier way to make it happen.

PR #38, opened 2018-10-30
Replace CDI extension with event observer

jpechane:


kstreams-live-update/aggregator/src/main/java/io/debezium/examples/kstreams/liveupdate/aggregator/StreamsPipelineManager.java
```diff
@@ -53,8 +52,7 @@
     private KafkaStreams streams;
     private ExecutorService executor;
 
-    @PostConstruct
-    public void startKStreams() {
+    public void startKStreams(@Observes @Initialized(ApplicationScoped.class) Object init) {
```
gunnarmorling: Ah, ok. I see now where I went wrong; instead passing `Object`,  I used `StreamsPipelineManager` as parameter type. But of course it's not the actual bean that's passed but instead it's the `ServletContext`. Hence my method wasn't invoked. Neat trick, didn't know that :)

gunnarmorling:
Applied. Thanks, @jpechane!

PR #42, opened 2018-11-05
[kstreams] Making aggregator a multi-stage Docker build;

gunnarmorling:
* One stage creates a minimal JDK with just the required modules
* One stage builds the project and adds the dependencies
* The final stage assembles all required files

As a result, the resulting image is much smaller.

### DIFF
--- a/kstreams/poc-ddd-aggregates/Dockerfile
+++ b/kstreams/poc-ddd-aggregates/Dockerfile
@@ -1,4 +1,41 @@
-FROM fabric8/java-jboss-openjdk8-jdk
+# 1. Prepare JDK
+FROM fedora:29 as jdk
 
-COPY target/poc-ddd-aggregates-0.1-SNAPSHOT.jar /deployments
-COPY target/dependencies/* /deployments/
+RUN mkdir /tmp/jdk \
+     && cd /tmp/jdk \
+     && curl -O https://download.java.net/java/GA/jdk11/13/GPL/openjdk-11.0.1_linux-x64_bin.tar.gz \
+     && tar -xvf openjdk-11.0.1_linux-x64_bin.tar.gz
+
+# Run this to get list of needed modules;
+# Not running it as part of the Dockerfile as it needs the app JAR as input, so it'd have to be executed each time
+# jdeps --print-module-deps --class-path poc-ddd-aggregates/target/dependencies/* poc-ddd-aggregates/target/poc-ddd-aggregates-0.1-SNAPSHOT.jar
+
+RUN /tmp/jdk/jdk-11.0.1/bin/jlink \
+    --add-modules java.desktop,java.management,java.naming,java.security.jgss,java.security.sasl,java.sql,jdk.unsupported \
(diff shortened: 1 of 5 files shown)


gunnarmorling:
Hey @jpechane, this is the outcome of a small weekend experiment :)

Of course it's not needed per se for this example, but I was curious how a Docker build could be structured so to take advantage of a smaller base image and the modular runtime images (subsets of the JDK) as provided by Java 9 and later.

I also did in preparation of future editions of the microservices lab, because the dependency retrieval can easily be extracted into its own image, which would then serve as base for the build stage. Then, in OpenShift, we'd simply use the Docker build strategy instead of S2I. This should avoid all the hassle we had with the pre-baked S2I image, and overall the created images are *much* smaller (~180 MB), meaning deployment will be quicker and workshop experience a tad smoother.

kstreams/poc-ddd-aggregates/Dockerfile
```diff
@@ -1,4 +1,41 @@
...
+RUN mkdir /tmp/jdk \
+     && cd /tmp/jdk \
+     && curl -O https://download.java.net/java/GA/jdk11/13/GPL/openjdk-11.0.1_linux-x64_bin.tar.gz \
+     && tar -xvf openjdk-11.0.1_linux-x64_bin.tar.gz
+
+# Run this to get list of needed modules;
+# Not running it as part of the Dockerfile as it needs the app JAR as input, so it'd have to be executed each time
+# jdeps --print-module-deps --class-path poc-ddd-aggregates/target/dependencies/* poc-ddd-aggregates/target/poc-ddd-aggregates-0.1-SNAPSHOT.jar
+
+RUN /tmp/jdk/jdk-11.0.1/bin/jlink \
```
jpechane: Maybe it is not necessary to include all VMs

gunnarmorling: WDYM by "all VMs"?

jpechane:
@gunnarmorling This is definitely neat idea. The only pitfall I found is if you touch the pom file (just adding a new config property) - then all the dependencies are completely redownloaded again.

kstreams/poc-ddd-aggregates/Dockerfile
jpechane: `--vm={client|server|minimal|all}`, default is `all` so to include all three

gunnarmorling:
> The  only pitfall I found is if you touch the pom file (just adding a new  config property) - then all the dependencies are completely redownloaded  again.

Yes, that's true. But I think that's acceptable in practice. In a way it indicates that it'd be nice if Maven would allow us to separate dependency info from all the other project metadata, but well... :)

kstreams/poc-ddd-aggregates/Dockerfile
gunnarmorling: Ah yes. Just tried it, but doesn't make a difference at all. I think this is because there are only the files of the server VM to begin with (if I set the option to "client" or "minimal", an exception is raised about that VM not existing).

jpechane:
@gunnarmorling Applied, thanks!

PR #43, opened 2018-11-06
[metrics] Metrics exposed via Prometheus/Grafana

jpechane:
To try this you need a custom 0.9 image with baked DBZ-776.
This is more of a first cut. Start via `docker-compose up --build`
Grafana is running at `http://localhost:3000`. there is pre-configured simple Debezium dashboard that expects SqlServer plugin running.
We can confirm that it works from the begining till end.
What should be done
1. Paramterize dashboard for arbitrary connector type and name
2. Define which metrics should be presented and in what form. Should we also include some of JVM metrics?


gunnarmorling:
> Paramterize dashboard for arbitrary connector type and name
+1

> Define which metrics should be presented and in what form. Should we also include some of JVM metrics?

I'd leave JVM metrics out for now. People can get hold of them in other ways, e.g. via the Strimzi dashboard.

gunnarmorling:
So it seems this needs a bit more work. I only see the "Time since last event" and "Events received" metricts.

jpechane:
@gunnarmorling This is now completed. Requires an image with https://github.com/debezium/debezium/pull/704 and https://github.com/debezium/debezium-incubator/pull/51

monitoring/docker-compose.yml
```diff
@@ -0,0 +1,57 @@
...
+     - kafka
+     - sqlserver
+    environment:
+     - BOOTSTRAP_SERVERS=kafka:9092
+     - GROUP_ID=1
+     - CONFIG_STORAGE_TOPIC=my_connect_configs
+     - OFFSET_STORAGE_TOPIC=my_connect_offsets
+     - STATUS_STORAGE_TOPIC=my_connect_statuses
+     - KAFKA_OPTS=-javaagent:/kafka/jmx_prometheus_javaagent.jar=8080:/kafka/config.yml
+     - JMX_PORT=1976
```
gunnarmorling: Who is the consumer of this env var? For Debezium's own [JMX configuration](https://github.com/debezium/docker-images/blob/master/connect-base/0.9/docker-entrypoint.sh#L78-L83), it'd have to be `JMXPORT` (no underscore), so I'm trying to figure where that is handled.

gunnarmorling:
@jpechane I've been thinking a bit about the issue with the string "metrics" (which aren't true metrics in Prometheus' world). How about exposing this as tabular data like so:

LastEvents

| Label | Value |
| Update xyz | 1 |
| Insert xyz | 2 |
| Update xyz | 3 |
| Update xyz | 4 |
| Update xyz | 5 |

(Instead of the counter 1 - n, this could also be the time in ms passed since then). And similarly for the captured tables, probably just with counter values from 1 - n. WDYT?

monitoring/docker-compose.yml
jpechane: `JMX-PORT` is used by Kafka scripts - see `kafka-run-class.sh`. I'd say that Debezium's option is superfluous.

jpechane:
@gunnarmorling The `time since last events` should be fixed now

jpechane:
@gunnarmorling How about storing them in a map where key is an offset and the value is the event itself? This should gurantee unqiuness of the key. And the map itself also needs to support evicition which we already have.

monitoring/debezium-grafana/debezium-dashboard.json
```diff
@@ -112,7 +112,7 @@
           "value": "null"
         }
       ],
-      "valueName": "avg"
+      "valueName": "current"
```
gunnarmorling: Nice one :)

monitoring/docker-compose.yml
gunnarmorling: Interesting; wondering why it was added then. Perhaps it didn't exist in KC itself back then? Anyways, should be good.

gunnarmorling:
Applied. Very nice, @jpechane! Also looking forward to further additions to that, e.g. lag and other things.

PR #44, opened 2018-11-21
DBZ-999

gunnarmorling:
https://issues.jboss.org/browse/DBZ-999

@jpechane This adds the KSQL example set-up. ~Still need to update the blog post in a few places (starting CLI).~


jpechane:
@gunnarmorling Applied, thanks!

PR #46, opened 2018-12-03
DBZ-991 Example for invalidating JPA 2nd-level cache after DB updates

gunnarmorling:
https://issues.jboss.org/browse/DBZ-991

Hey @jpechane, check out this one. It uses the embedded engine to evict items from the 2nd-level cache after doing data updates bypassing the application. In order to not invalidate items altered by the application itself, it uses a small Hibernate ORM extension, which keeps track of all transactions run by the app itself, so that the DBZ event handler can exclude any events pertaining to those transactions.

I'll write a blog post next.


jpechane:
@gunnarmorling Looks very good, there is not much to complain about. I have just two proposals
1) Replace Hibernate-only integrator with standard CDI-based observer - http://docs.jboss.org/cdi/spec/2.0/cdi-spec.html#transactional_observer_methods
2) One of the steps in the demo might be to show that entity without `@Cacheable` annotation is not affected

Also it might be useful to discuss in the text the eventual consistency of this solution.

It s little bit sad that Inifinispan in library mode is not used as known tx cache.

One idea for detection of in-app out-of-app changes - if optimistic-licking is enabled for the entity then when the version fild is changed we can suppose it is done by application, otherwise it is done extrenally.

gunnarmorling:
> Replace Hibernate-only integrator with standard CDI-based observer - http://docs.jboss.org/cdi/spec/2.0/cdi-spec.html#transactional_observer_methods

I liked the idea, but as discussed it requires to emit an event from the app which isn't desirable. One potential way out could be to observe the `RequestScope` and do it transparently that way. But for now I'll stick to the extension to ORM.

> Also it might be useful to discuss in the text the eventual consistency of this solution.

Good idea, done.

> It s little bit sad that Inifinispan in library mode is not used as known tx cache.

Grrr, good idea which cost me only one hour or so :) Done.

> One idea for detection of in-app out-of-app changes - if optimistic-licking is enabled for the entity then when the version fild is changed we can suppose it is done by application, otherwise it is done extrenally.

Not sure whether I'm following on this one. How would the "is changed" look like exactly? In any case it's limited to a specific subset of models (those using optimistic locking), so I'll stick to the original approach.

Thanks a lot for reviewing!

PR #52, opened 2019-02-01
outbox pattern

gunnarmorling:
Hey @jpechane, I still need to add the consumer side, but I think this can go in already. Obviously, things will be more clear with the blog post. I'll add some comments inline to point out interesting bits. Feedback welcome!


outbox/order-service/src/main/java/io/debezium/examples/outbox/order/event/ExportedEvent.java
```diff
@@ -0,0 +1,19 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.examples.outbox.order.event;
+
+import com.fasterxml.jackson.databind.JsonNode;
+
+public interface ExportedEvent {
```
gunnarmorling: That describes events in outbox table.

outbox/order-service/src/main/java/io/debezium/examples/outbox/order/outbox/EventSender.java
```diff
@@ -0,0 +1,42 @@
...
+    @Inject
+    private Event<OutboxEventSubmitted> submitted;
+
+    @Inject
+    private HouseKeeper houseKeeper;
+
+    @PersistenceContext
+    private EntityManager entityManager;
+
+    public void onExportedEvent(@Observes ExportedEvent event) {
```
gunnarmorling: Create an entry in the outbox able when receiving an (synchronous, in-TX CDI event).

outbox/order-service/src/main/java/io/debezium/examples/outbox/order/outbox/EventSender.java
```diff
@@ -0,0 +1,42 @@
...
+        OutboxEvent outboxEvent = new OutboxEvent(
+                event.getAggregateType(), event.getAggregateId(), event.getType(), event.getPayload()
+        );
+        entityManager.persist(outboxEvent);
+
+        submitted.fire(new OutboxEventSubmitted(outboxEvent.getId()));
+    }
+
+    public void onEventSubmitted(@Observes(during=TransactionPhase.AFTER_COMPLETION) OutboxEventSubmitted event) {
```
gunnarmorling: And remove the entry right away. That way the event table is always empty, we'll get the new events from the log.

jpechane: Should not it be `AFTER_SUCCESS`?  When tx fails there should be no record in the table.

outbox/order-service/src/main/java/io/debezium/examples/outbox/order/outbox/EventSender.java
```diff
@@ -0,0 +1,42 @@
...
+    @PersistenceContext
+    private EntityManager entityManager;
+
+    public void onExportedEvent(@Observes ExportedEvent event) {
+        OutboxEvent outboxEvent = new OutboxEvent(
+                event.getAggregateType(), event.getAggregateId(), event.getType(), event.getPayload()
+        );
+        entityManager.persist(outboxEvent);
+
+        submitted.fire(new OutboxEventSubmitted(outboxEvent.getId()));
```
jpechane: Have you tried `INSERT` and `DELETE` in the same transaction? I belive it should still create two events and the table will be kept empty and code will be simplified.

outbox/order-service/src/main/java/io/debezium/examples/outbox/order/outbox/EventSender.java
gunnarmorling: Ah yes, good point.

outbox/order-service/src/main/java/io/debezium/examples/outbox/order/outbox/EventSender.java
gunnarmorling: Good question, too. So Hibernate itself would optimize that and just do nothing. But what actually could work is to do this: insert - flush - delete. I'll try that.

gunnarmorling: Ok, so I changed that.

> So Hibernate itself would optimize that and just do nothing

Turns out I was actually just dreaming that :) When doing `persist()` + `remove()` right after each other, it will produce an INSERT followed by DELETE. Steve confirmed there's no optimization for that. Next thing I wondered whether those could be omitted by the DB in the TX log, but that's not the case either. So it's all much simpler now indeed!

gunnarmorling:
@jpechane Added the consuming service, too. Can you take a look at the duplicate check in particular? What I don't like is that it requires to fully parse the incoming JSON payload. This could be avoided if the event UUID was passed as header. I can do that in the routing SMT, but the CDI extension doesn't expose headers to consumers unfortunately, so I'm considering to fall back to a regular Kafka consumer. /CC @matzew

outbox/event-routing-smt/src/main/java/io/debezium/examples/outbox/routingsmt/EventRouter.java
```diff
@@ -0,0 +1,84 @@
...
+    public EventRouter() {
+
+    }
+    @Override
+    public void configure(Map<String, ?> configs) {
+    }
+
+    @Override
+    public R apply(R record) {
+        Struct struct = (Struct) record.value();
```
jpechane: I think here should be a condiiton against tombstone.

outbox/order-service/src/main/java/io/debezium/examples/outbox/order/outbox/EventSender.java
```diff
@@ -0,0 +1,27 @@
...
+
+    @PersistenceContext
+    private EntityManager entityManager;
+
+    public void onExportedEvent(@Observes ExportedEvent event) {
+        OutboxEvent outboxEvent = new OutboxEvent(
+                event.getAggregateType(), event.getAggregateId(), event.getType(), event.getPayload()
+        );
+
+        entityManager.persist(outboxEvent);
```
jpechane: @gunnarmorling This is much easier but I think it deserves a collocated comment.

outbox/shipment-service/src/main/java/io/debezium/examples/outbox/shipment/log/MessageLog.java
```diff
@@ -0,0 +1,32 @@
...
+import java.util.UUID;
+
+import javax.enterprise.context.ApplicationScoped;
+import javax.persistence.EntityManager;
+import javax.persistence.PersistenceContext;
+import javax.transaction.Transactional;
+import javax.transaction.Transactional.TxType;
+
+@ApplicationScoped
+public class MessageLog {
```
jpechane: @gunnarmorling This could be nice use case for ISPN tx cache

jpechane:
@gunnarmorling LGTM, I've add just few suggestions

outbox/order-service/src/main/java/io/debezium/examples/outbox/order/outbox/EventSender.java
gunnarmorling: Will do.

outbox/shipment-service/src/main/java/io/debezium/examples/outbox/shipment/log/MessageLog.java
gunnarmorling: I had thought about that, too, but decided against it as it adds more complexity again:

* needs persistent cache store so its data is still present after restarts
* needs to be transactional, so that consistency with the app's DB is ensured; in the current approach, e.g. the log record will be rolled back if the business op failed for some reason, just by virtue of using the same DB. Avoiding 2PC is kind of the reason for the entire set-up to begin with, so I think using the DB is preferable.

jpechane: @gunnarmorling I understand, then I propose to add a business table to the database and demo so the benefits are obvious on the first sight.

gunnarmorling: Ok, I can do that.

outbox/event-routing-smt/src/main/java/io/debezium/examples/outbox/routingsmt/EventRouter.java
gunnarmorling: Done.

outbox/order-service/src/main/java/io/debezium/examples/outbox/order/outbox/EventSender.java
gunnarmorling: Done.

outbox/shipment-service/src/main/java/io/debezium/examples/outbox/shipment/log/MessageLog.java
gunnarmorling: Done.

PR #54, opened 2019-02-15
Do a null check, because the value node is null for the deletes.

gnomezgrave:
**Type**: Bug
**Severity**: Critical

When there are delete records coming in, this program crashes because the `value` is null. So, we first need to check if the `value` (or `valueSchema`) is null, before using the value.


gunnarmorling:
Good spot, thanks! Merging.

PR #55, opened 2019-03-08
Non-native variant of Quarkus based outbox

jpechane:


jpechane:
This is a first cut of outbox with shipment service based upon Quarkus. Main changes done
 * Quarkus does not support MySQL yet, so MariaDB is used instead of it
 * `KafkaConsumer` observes application scope destroyed event right after start immediatelly closing the consumer. The code has been temproarily disabled and will probably be replace with shutdown hook

Native image has not been tried yet.

gunnarmorling:
That's a great start, thanks a lot, @jpechane!

I'd suggest three things:

* Instead of copying the entire example, simply add _shipment-service-quarkus_ to the existing set-up, using a different consumer group
* Use the SmallRye Reactive Messaging API for consuming messages (see [the quickstart](https://github.com/quarkusio/quarkus-quickstarts/tree/master/kafka-quickstart) for an example)
* See what it takes to go native

gunnarmorling:
> observes application scope destroyed event right after start immediatelly closing the consumer.

On that one, seems like it could be a bug with the Quarkus CDI implementation (Arc)? Could you open a bug report upstream for it, please?

jpechane:
@gunnarmorling All comments incorporated, there are now two shipping services, the Quarkus based is using native image. Each service has its own db.
There is only one half-open issue - as I mentioned the Quarkus based services starts too fast. To make it production grade we should include a healthcheck REST endpoint that would verify validity of service componenets and report it back so Docker compose can use helthcheck to retart it in case of failure.
Right now I added just a 5 secs sleep to the compose command so the db has a chance to start.
It is now a question whether we want to add the infrasturcture code to the demo or not.

gunnarmorling:
@jpechane Excellent! On the timing issue, did you take a look at the healthcheck implementation of the _kstreams-live-update_ demo? This causes the app container to wait until the DB container is in "healthy" state, which is the case if "Select 1" or so returns successfully.

gunnarmorling:
> there are now two shipping services

I still see files under _outbox-quarkus_, whereas I would have expected new things only under _outbox/shipment-service-quarkus_?

jpechane:
@gunnarmorling I moved the service but forgot to delete the dir. You can ignore it now completely, I'll push the removal when I get to the computer. You are right with the probe, I'd added it to db instead of app container and it did not work - late in the day conclusions :-(

gunnarmorling:
Ok, cool. So it seems these are the differences to the original implementation:

* Uses SmallRye Reactive Messaging for message consumption instead of Kafka client API
* Uses Jackson instead of JSON-P
* Uses MariaDB instead of MySQL

Anything else? If that's it, I think it'd be great to have a blog post describing these steps needed for "migration", the required build steps (calling Maven with the native profile or whatever it is) and also outline the better numbers for start-up and memory usage.

outbox/README.md
```diff
@@ -10,7 +10,7 @@ excluding any duplicate messages by comparing incoming event ids with already su
 
 Prepare the Java components:
 
-    mvn clean install
+    mvn clean install -Pnative -Dnative-image.docker-build=true
```
gunnarmorling: So I reckon this requires Graal to be installed locally, right? How about pushing the build into a (multi-stage) Docker build, simplifying that set-up: https://github.com/debezium/debezium-examples/blob/master/kstreams/poc-ddd-aggregates/Dockerfile

gunnarmorling:
Squashed and merged. Thanks a lot, @jpechane! Added one more commit for some clean-up. Looking forward to the blog post!

jpechane:
https://github.com/quarkusio/quarkus/issues/1433

PR #58, opened 2019-03-16
Upgrading examples to Debezium 0.9 and other minor typo fixes

opheliagame:


end-to-end-demo/debezium-jdbc/Dockerfile
```diff
@@ -1,4 +1,6 @@
-FROM debezium/connect:0.8
+ARG DEBEZIUM_VERSION=0.9
```
vandaei: Good call. Yes I also lost hours due to hard-coded value in this file to 0.8. Setting up Oracle example I was changing the version to 0.9 in .yaml file knowing 0.8 has issues with Oracle but yet I was having problem. Finally figured out there is a hard coded version here for connect core. I have taken a bit different approach to make it fully configurable. #61

gunnarmorling:
Thanks a lot, @opheliagame; great improvements!

PR #60, opened 2019-03-25
[DBZ-1169] Update Outbox example

renatomefi:
## Context

With Outbox now part of Debezium core the example can be update
https://issues.jboss.org/browse/DBZ-1169

This is **Work in Progress**

### Notice

Since the new Debezium `0.9` Docker images aren't pushed you have to build them locally first

## Todo

- [x] Create a timestamp field in the table to influence the event time
- [x] Update README

## Awesome
Screenshot from 2019-03-26 13-02-24


outbox/event-routing-smt/src/main/java/io/debezium/examples/outbox/routingsmt/EventRouter.java
```diff
@@ -1,88 +0,0 @@
-/*
```
gunnarmorling: Nice, nothing beats deleting code. Great sign that things are getting easier right out-of-the-box!

outbox/order-service/src/main/java/io/debezium/examples/outbox/order/outbox/ExportedEvent.java
```diff
@@ -35,4 +35,9 @@
      * The type of an event, e.g. "Order Created", "Order Line Canceled" for events of the "order" aggregate type.
      */
     String getType();
+
+    /**
+     * The timestamp which the event has happened
```
gunnarmorling:
```suggestion
     * The timestamp at which the event has happened
```

outbox/register-postgres.json
```diff
@@ -1,18 +1,16 @@
...
+    "database.hostname": "order-db",
+    "database.port": "5432",
+    "database.user": "postgresuser",
+    "database.password": "postgrespw",
+    "database.dbname" : "orderdb",
+    "database.server.name": "dbserver1",
+    "schema.whitelist": "inventory",
+    "table.whitelist" : "inventory.outboxevent",
+    "tombstones.on.delete" : "false",
+    "transforms" : "outbox",
```
gunnarmorling: Wow, that's freakin' cool to see how well it works just without much config 👍

outbox/shipment-service-quarkus/src/main/resources/application.properties
```diff
@@ -5,7 +5,7 @@ quarkus.datasource.password: mariadbpw
 quarkus.hibernate-orm.database.generation=drop-and-create
 quarkus.hibernate-orm.log.sql=true
 smallrye.messaging.source.orders.type=io.smallrye.reactive.messaging.kafka.Kafka
-smallrye.messaging.source.orders.topic=OrderEvents
+smallrye.messaging.source.orders.topic=outbox.event.order
```
gunnarmorling: Might be worth showing how to actually keep the more custom topic name?

outbox/README.md
```diff
@@ -16,59 +16,83 @@ This allows to build a native binary of that service, resulting in significantly
...
+```console
+$ export DEBEZIUM_VERSION=0.9
+$ docker-compose up --build
+```
 
 Register the Debezium Postgres connector:
 
-    cat register-postgres.json  | http POST http://localhost:8083/connectors/
+```console
+$ http PUT http://localhost:8083/connectors/outbox-connector/config < register-postgres.json
```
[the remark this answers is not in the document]
renatomefi: PUT is an idempotent operation, you can decide the name of the connector in the url and always post a new configuration, independently if it was created before or not. I tend to use this one because the `json` payload is uniform, if you want to do updates in the connector you only need a single payload now, instead of two

outbox/event-routing-smt/src/main/java/io/debezium/examples/outbox/routingsmt/EventRouter.java
renatomefi: Feels good right?! :smile:

outbox/shipment-service-quarkus/src/main/resources/application.properties
renatomefi: I might have screwed this up, so I'm lowering case the aggregate type, so I couldn't find a way to do the regex replacement while upper casing the letter `E` for `OrderEvents`, what I could do is put the topic name to `order.events` for instance or work on a fix for the outbox to be more permissive, but then I don't know how to lower case it later as well! Wdyt?

outbox/README.md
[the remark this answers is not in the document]
gunnarmorling: Gotcha. Didn't know one could use PUT for creation, too. Thanks for sharing, I'll adopt that :)

outbox/shipment-service-quarkus/src/main/resources/application.properties
gunnarmorling: Would it work to use "Order" as aggregate type, and then "${routedByValue}Events" for the `route.topic.replacement` SM option?

renatomefi: The final result with this is `orderEvents` due to the lower casing in the transform, I commited so you can see how it looks like!

renatomefi:
Hey @gunnarmorling, so sorry for the delay!

> Maybe we could have the blog post we discussed online next week?


Yes, I really want to make this blogpost, I'll make a draft PR soon and you can guide me!

outbox/shipment-service-quarkus/src/main/resources/application.properties
gunnarmorling: Thanks! Where is the lower-casing coming from? I think I'm still missing a piece of the puzzle to understand.

gunnarmorling:
> Yes, I really want to make this blogpost, I'll make a draft PR soon and you can guide me!

Excellent, looking forward to it very much. In a nutshell, it'd have to be a new file [here](https://github.com/debezium/debezium.github.io/tree/develop/blog). You should also add yourself as an author [here](https://github.com/debezium/debezium.github.io/blob/develop/_config/authors.yml), so you show up on the rendered page. The [README](https://github.com/debezium/debezium.github.io/blob/develop/README.md) has all the other details; thankfully, it's all Docker-based, so you don't need to fiddle with any local installs of the blogging engine. Happy to help as needed. Thanks again!

outbox/shipment-service-quarkus/src/main/resources/application.properties
renatomefi: I introduced it here, my line of thought was that I "assumed" that often aggregate types would be camel case, and that would force it to the topic naming, but I actually didn't think that people could be using upper case in their topic names, since what I usually saw is lowercase (including kafka connect official topics, consumer offsets, and such), so I made it to lowercase.
So now that I see it was a bunch of assumptions I want your input on what can be a proper way to deal with it, I tried to find a solution using the regex but couldn't, so not sure if I should just pass it through or make it configurable.
Since it's a dynamic router I'm not sure if even with another router I could achieve it nicely

gunnarmorling: > since what I usually saw is lowercase (including kafka connect official topics, consumer offsets, and such),

I see, that makes sense. Indeed I don't think one can do any changes to casing using Java regexps. So how about going for "order.events" as topic name then (and adjust the aggregate name used by the producer to "order" if needed?

gunnarmorling:
Hey @renatomefi, just checking in on this one. Anything we should talk about?

gunnarmorling:
And applied. Thanks a lot, @renatomefi! Pushed one more commit for a few README tweaks. Thanks for your patience around the topic name dicussion. It looks really good now 🎉  The only remaining thing would be the blog post we discussed (I know, I'm greedy... ;)

PR #61, opened 2019-03-26
remove hardcoded version parameter

vandaei:
To make it configurable same as docker-copose<connector>.yaml
Oracle is fixed in version 0.9 and is not working for version 0.8 and I was expecting by setting export DEBEZIUM_VERSION=0.9 in shoul get fixed but in fact all other components (kafka and zookeeper) were getting on 0.9 but the connect was yet 0.8 although image was: image: debezium/connect-with-oracle-jdbc:${DEBEZIUM_VERSION}


opheliagame:
That's definitely more flexible! Good thinking :)

gunnarmorling:
Good catch, thanks a lot, @vandaei!

PR #62, opened 2019-04-17
Kstreams quarkus

gunnarmorling:
@jpechane, @Naros, this is part of my Summit preparations - migrating the existing KStreams demo to Quarkus. For now HotSpot only, I'll try whether KStreams actually can be made working under SVM, that'd be pretty cool, too :)


gunnarmorling:
Merging.

PR #63, opened 2019-05-24
Fixed error in docker build where no file matched regex for copy (src)

zacyang:


gunnarmorling:
Good catch, thanks! Weird that it slipped through so far.

PR #64, opened 2019-06-26
DBZ-1359 Debezium to Apache Camel example

jpechane:
https://issues.jboss.org/browse/DBZ-1359


gunnarmorling:
Oh, and could we make this a Quarkus app perhaps, running in native mode? It does have Camel support AFAIK. That'd also spice it up some more.

jpechane:
@gunnarmorling Please check https://camel.apache.org/components.html for thje list of supported components.
We need to balance between excitement and setup. I think we can do either solr or etcd or if we put more work on user it can be Twitter/Dropbxo/github but that would require the user to privde an account.

gunnarmorling:
Salesforce actually is a very interesting one, as it's a realistic and important use case. But I agree it'll be (too) hard to set it up.

Something around Twitter perhaps, although I'm not having a great idea for an actual application of it right now. Perhaps this: a Q&A site such as SO wishes to post a tweet with a link whenever there's a new answer to a question; or if an answer's rating goes beyond 10 upvotes or so (the latter might show some custom filtering in Camel). WDYT?

jpechane:
@gunnarmorling How about having a table with alerts/errors? When a record is create din it it is detected in route and a message from one of the fields is delivered to a Twitter account?

gunnarmorling:
@jpechane It seems a bit contrived; in which use case would such alerts be stored in a DB? Wouldn't one rather directly push a message to Kafka from the application that raises the alert? That's what I like about the Q&A example: there it really makes sense to react to a data change.

jpechane:
This PR will be updated when Camel 3 RC2 will be released - see https://jira.apache.org/jira/browse/CAMEL-12543

gunnarmorling:
Closing this in favour of #88.

PR #66, opened 2019-08-22
[auditlog] Initial import

gunnarmorling:
Hey @Naros, here's the first cut of the demo we've been talking about earlier today. Check out the README.md for an overview. The most interesting parts are `TransactionInterceptor` in the _vegetables-service_ which does the insert into the audit metadata table and `TopologyProducer` which does the enrichment of change events based on that metadata via KStreams.

Happy about any feedback from your side :) Thanks!


auditlog/vegetables-service/src/main/java/io/debezium/demos/auditing/vegetables/transactioncontext/TransactionContextData.java
```diff
@@ -0,0 +1,25 @@
...
+    @Column(name="transaction_id")
+    public long transactionId;
+
+    @Column(name="user_name")
+    public String userName;
+
+    @Column(name="client_date")
+    public ZonedDateTime clientDate;
+
+    public String usecase;
```
hpgrahsl: nitpick -> shouldn't that also be `useCase` and then have an annotation for "use_case" as well to be consistent with naming?

auditlog/docker-compose.yaml
```diff
@@ -0,0 +1,69 @@
...
+  kafka:
+    image: debezium/kafka:${DEBEZIUM_VERSION}
+    ports:
+     - 9092:9092
+    links:
+     - zookeeper
+    environment:
+     - ZOOKEEPER_CONNECT=zookeeper:2181
+     # For local development of auditlog-enricher
+     # - ADVERTISED_HOST_NAME=192.168.1.6
```
hpgrahsl: ah obviously some changes while I reviewed. yes this should be commented out or set to the correct IP address of the host machine from the very beginning ;-)

auditlog/log-enricher/src/main/java/io/debezium/demos/auditing/enricher/TopologyProducer.java
```diff
@@ -0,0 +1,71 @@
...
+                                                .add("audit", Json.createObjectBuilder(
+                                                        txData.get("after")
+                                                            .asJsonObject())
+                                                            .remove("transaction_id")
+                                                            .build()
+                                                )
+                                                .build())
+                                    .build()
+                )
+                // re-key by original PK
```
hpgrahsl: more or less the same as above so just using map, and I'd rename the params to `txId` and `enrichedEvent` cause it makes more sense to me when reading through the kstreams pipeline :) 

```java

.map((txId, enrichedEvent) -> KeyValue.pair(
                        enrichedEvent.get("id").asJsonObject(),
                        enrichedEvent.get("changeEvent").asJsonObject()
))

```

auditlog/log-enricher/src/main/java/io/debezium/demos/auditing/enricher/TopologyProducer.java
```diff
@@ -0,0 +1,71 @@
...
+    private static final String VEGETABLES_ENRICHED_TOPIC = "dbserver1.inventory.vegetable.enriched";
+
+    @Produces
+    public Topology buildTopology() {
+        StreamsBuilder builder = new StreamsBuilder();
+
+        KTable<JsonObject, JsonObject> transactionContextData = builder.table(TX_CONTEXT_DATA_TOPIC);
+
+        builder.<JsonObject, JsonObject>stream(VEGETABLES_TOPIC)
+                // store id in message value so we can restore it later on
```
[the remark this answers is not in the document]
gunnarmorling: Ah, yes. Of course. Thanks!

auditlog/log-enricher/pom.xml
```diff
@@ -0,0 +1,123 @@
...
+<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns="http://maven.apache.org/POM/4.0.0"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>io.debezium.demos.auditing</groupId>
+  <artifactId>auditing-log-enricher</artifactId>
+  <version>1.0-SNAPSHOT</version>
+  <properties>
+    <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
+    <surefire-plugin.version>2.22.0</surefire-plugin.version>
+    <!-- <quarkus.version>0.21.1</quarkus.version>  -->
```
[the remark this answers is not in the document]
gunnarmorling: Yeah,  I was using a locally built snapshot. Good catch, thanks!

auditlog/vegetables-service/src/main/java/io/debezium/demos/auditing/vegetables/transactioncontext/TransactionContextData.java
```diff
@@ -21,5 +21,6 @@
     @Column(name="client_date")
     public ZonedDateTime clientDate;
 
-    public String usecase;
+    @Column(name="use_date")
```
hpgrahsl: TYPO in annotation -> use_date ? ;-)

gunnarmorling: Uff, yes :) I could need more coffee it seems :)

gunnarmorling:
Pushed a commit to address the first batch of comments; thanks a lot, @hpgrahsl! Will look into the pipeline simplication now; biggest issue is handing of tombstones which currently still fails.

auditlog/log-enricher/src/main/java/io/debezium/demos/auditing/enricher/TopologyProducer.java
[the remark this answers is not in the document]
gunnarmorling: Done as well.

auditlog/README.md
```diff
@@ -0,0 +1,107 @@
...
+```
+
+This uses a pre-generated JWT token (with expiration date set to 2099-12-31 and user set to "farmerbob").
+To regenerate the token with different data, use the [Jwtenizr](https://github.com/AdamBien/jwtenizr) tool under _vegetables-service/jwt_.
+
+You can also update an existing vegetable record:
+
+```console
+$ http PUT http://localhost:8080/vegetables/10 'Authorization: Bearer eyJraWQiOiJqd3Qua2V5IiwidHlwIjoiSldUIiwiYWxnIjoiUlMyNTYifQ.eyJzdWIiOiJmYXJtZXJib2IiLCJ1cG4iOiJmYXJtZXJib2IiLCJhdXRoX3RpbWUiOjE1NjY0NTgxMTMsImlzcyI6ImZhcm1zaG9wIiwiZ3JvdXBzIjpbImZhcm1lcnMiLCJjdXN0b21lcnMiXSwiZXhwIjo0MTAyNDQ0Nzk5LCJpYXQiOjE1NjY0NTgxMTMsImp0aSI6IjQyIn0.CscbJN8amqKryYvnVO1184J8F67HN2iTEjVN2VOPodcnoeOd7_iQVKUjC3h-ye5apkJjvAsQKrjzlrGCHRfl-n6jC9F7IkOtjoWnJ4wQ9BBo1SAtPw_Czt1I_Ujm-Kb1p5-BWACCBCVVFgYZTWP_laz5JZS7dIvs6VqoNnw7A4VpA6iPfTVfYlNY3u86-k1FvEg_hW-N9Y9RuihMsPuTdpHK5xdjCrJiD0VJ7-0eRQ8RXpycHuHN4xfmV8MqXBYjYSYDOhbnYbdQVbf0YJoFFqfb75my5olN-97ITsi2MS62W_y-RNT0qZrbytqINA3fF3VQsSY6VcaqRAeygrKm_Q' 'Date: Thu, 22 Aug 2019 08:12:31 GMT' name=Pear description=tasty
```
Naros: See my comment above, same problem here with the headers and the trailing space.

gunnarmorling:
Hey @hpgrahsl, pushed an updated with the reworked as we discussed. It now doesn't do the join anymore to avoid issues in cases where order change event is processed before the TX metadata event has arrived.

I'm not super-happy with the implementation, but is the best I can think of right now. The look-up in the store happens in a loop, which of course could be done more nicely using Awaitility or similar. One also might think about using some kind of latches, but those would have to be hold in a map, so it might not be worth the effort and not even sure whether it's more efficient. Usually the TX record is found upon first or second attempt.

Feedback welcome!

auditlog/log-enricher/src/main/java/io/debezium/demos/auditing/enricher/TopologyProducer.java
```diff
@@ -0,0 +1,108 @@
...
+                if (metadata != null) {
+                    LOG.info("Found TX metadata after {} attempt(s)", i);
+                    return Json.createObjectBuilder(metadata.get("after").asJsonObject())
+                            .remove("transaction_id")
+                            .build();
+                }
+                i++;
+
+                try {
+                    Thread.sleep(1);
```
hashhar: This will cause your consumer to be rebalanced since you are sleeping for > MAX_POLL_MS on the main thread.

Try looking into the `punctuate()` method of the Processor or Transformer API.

gunnarmorling: Hey @hashhar, thanks a lot for giving this a look! Interesting remark with the rebalance; seems there's no way around using `punctuate()` then. What bothers me most about this: I may have to buffer multiple messages with the same key in the stream-side statestore (e.g. two UPDATEs for the same purchase order, whose TX event hasn't arrived yet. Is there any support for that in the state store API (multi-value for a given key), or would I have to collect the events in my own value holder (a JSON array essentially)?

hpgrahsl: @gunnarmorling as far as the state stores are concerned they operate on Bytes(key) and byte[] (value) see e.g. the RocksDB or InMem implementations. This essentially means you either provide De/Serializers for custom data structures (e.g. List of something) or go with JSON Serdes like you suggest.

gunnarmorling: Ok; so this all is getting much more complex than I had anticipated, but it's interesting, too :)

How about the following design/algorithm then:

* have a custom K/V store managed by a stream-side `Processor` ("buffer")
* have a punctuate method on that processor using wall-clock time

* when a stream event (purchase order) arrives
  * when the buffer is not empty
    * process buffered elements in insertion order
      * lookup TX metadata in GlobalKTable store
      * when found -> emit joined record
      * when not found, abort loop
    * handle stream event as in "buffer is empty case"
  * when the buffer is empty
    * lookup TX metadata in GlobalKTable store
      * when found -> emit joined record
      * when not found -> add stream event to buffer
* when punctuate is called
  * when the buffer is not empty: same logic as above

IIUC, there shouldn't be the need for `punctuate()` on the table side of the join, as the wall-clock punctuation on the stream side will process the buffer based on the latest additions to the `GlobalKTable`.

gunnarmorling: @renatomefi, as we were talking about using punctuation before, it seems that's the way to go as per @hashhar's comment above. I remember you mentioned you'd be interested in exploring this route, WDYT about the approach above? Let me know if you're interested, I can also wrap this up (next week likely).

gunnarmorling:
Hey @hpgrahsl, @hashhar, @renatomefi and @jpechane, I've pushed one more commit which eventually I believe is the right way™ for implementing the waiting logic. It's quite simple actually:

* when receiving a change event on the stream side of the join whose associated TX event cannot be found yet, that event will be added to key/value store which serves as a buffer
* before processing the incoming event, all buffered events will be tried to be processed in their insertion order; if that's successful, the incoming event will be processed, otherwise it will be buffered
* via punctuation the buffer is regularly processed, too

Note one could think about processing the buffer when receiving an event at the table-side, too (so to reduce the lag in case no stream-side event arrives, in which case only the next punctuation call would pick it up), but I omitted that for the sake of simplicity.

Any feedback and testing will be welcomed. I'll focus on writing up my findings now :) Thanks!

hpgrahsl:
@gunnarmorling very nice solution for the above (and also on Slack) discussed challenges. as briefly written in a side-channel conversation this looks very reasonable and looks fine from what I saw during reading the code. however I couldn't run it yet... will probably try it out tomorrow.

auditlog/log-enricher/src/main/java/io/debezium/demos/auditing/enricher/TopologyProducer.java
```diff
@@ -0,0 +1,280 @@
...
+            builder.addStateStore(streamBufferStateStore);
+
+        builder.globalTable(txContextDataTopic, Materialized.as(STORE_NAME));
+
+        builder.<JsonObject, JsonObject>stream(vegetablesTopic)
+                // filter out any tombstones; for an audit log topic, time-based retention
+                // seems more reasonable than compaction
+                .filter((id, changeEvent) -> changeEvent != null)
+                // exclude snapshot events
+                .filter((id, changeEvent) -> !changeEvent.getString("op").equals("r"))
```
hashhar: Why are snapshot events ignored? I'm unable to see any reason why the existing code would fail to work for them.

hashhar:
@gunnarmorling The code for the Transformer looks correct and I can't see any issues with it. Will be doing a deep dive over the weekend into other areas of the code.

Very nice implementation with the `BufferOffsets` class.

auditlog/log-enricher/src/main/java/io/debezium/demos/auditing/enricher/TopologyProducer.java
gunnarmorling: Good question, two things basically:

* As snapshot events are *read* events, I didn't see much reason to have them in an audit log kind of topic
* A bit more of an actual issue is that for snapshot events there won't be any corresponding TX record issued by the application typically; the snapshot TX is initiated by the Debezium connector, so the user would have to go, retrieve the TX id from one of the events and produce a matching record into the TX metadata topic; otherwise the snapshot events would block any other log-based events from being processed

So based on that I figured that's the best for the demo; in the post I'll discuss it, also showing options, e.g. one could want to propagate snapshot events un-enriched if that makes sense in a given use case.

gunnarmorling:
Thanks for reviewing, @hpgrahsl and @hashhar! Very glad to hear you don't see any obvious issues with the suggested implementation. Took definitely longer than I had anticipated :)

auditlog/log-enricher/src/main/java/io/debezium/demos/auditing/enricher/TopologyProducer.java
```diff
@@ -36,73 +46,235 @@
...
+        private KeyValueStore<Long, JsonObject> streamBuffer;
 
         @Override
         @SuppressWarnings("unchecked")
         public void init(ProcessorContext context) {
+            this.context = context;
+            streamBuffer = (KeyValueStore<Long, JsonObject>) context.getStateStore(STREAM_BUFFER_NAME);
             txMetaDataStore = (KeyValueStore<JsonObject, JsonObject>) context.getStateStore(STORE_NAME);
+
+            context.schedule(Duration.ofSeconds(1), PunctuationType.WALL_CLOCK_TIME, ts -> {
```
hpgrahsl: nitpick: no need for block lambda ;-)

auditlog/log-enricher/src/main/java/io/debezium/demos/auditing/enricher/TopologyProducer.java
```diff
@@ -36,73 +46,235 @@
...
+        private void bufferChangeEvent(JsonObject key, JsonObject changeEvent) {
+            LOG.info("Buffering change event for key {}", key);
+
+            BufferOffsets sequence = bufferOffsets().orElseGet(BufferOffsets::initial);
+
+            JsonObject wrapper = Json.createObjectBuilder()
+                    .add("key", key)
+                    .add("changeEvent", changeEvent)
+                    .build();
+            streamBuffer.put(sequence.getNextValueAndIncrement(), wrapper);
```
hpgrahsl: You might want to "batch" these two consecutive put-calls together. You simply use putAll Method for that and hand it a List of both entries. The underlying RocksDB store applies the batch atomically then. WDYT?

auditlog/log-enricher/src/main/java/io/debezium/demos/auditing/enricher/TopologyProducer.java
```diff
@@ -0,0 +1,280 @@
...
+    /**
+     * Keeps track of the position within the key/value state store that is used as
+     * a buffer for non-joinable events. As there's no true queue like structure
+     * offered by Kafka Streams, a statestore is used, whose keys indicate the order
+     * of insertion and whose value are the buffered change events. This structure
+     * keeps track of the oldest insert position and the next insert position. It is
+     * stored in the same state store using a special key outside of the sequence
+     * range.
+     */
+    private static class BufferOffsets {
```
jpechane: @gunnarmorling This is very clever solution but I believe it complicates the example.
Looking at the code I can firmly claim that the store's iterator is ordered, the only problem is that the ordering is lexicographic by the key.
So to make it a little bit more idiomatic I propose to
 - introduce a logical timestamp or use a record metadata timestamp extractor to associate order id with each event
- use the keystore where the key will be the timestamp from above converted into string and left padded
- use get/set/all/range for keystore as needed as the order is guranteed

gunnarmorling: @jpechane, that'd mean to rely on specifics of the `KeyValueStore` implementation, which isn't adviseable. The [docs](https://kafka.apache.org/23/javadoc/org/apache/kafka/streams/state/ReadOnlyKeyValueStore.html#all--) of `all()` are very clear about this:

> No ordering guarantees are provided.

So things might work now, but may fail when using a different store implementation or when its implementation gets changed in a future version. Now one could do a *custom* statestore which provides the required ordering guarantees, but I think that'd even more complex than what's suggested here.

jpechane: @gunnarmorling I am sorry for being pesky but as I read the code you've written a queue implementation on top of keystore. Would not it be possible to use Kafka API and make a buffer in a topic?
So the semantics will be the same
 - new message arrives
 - is buffer empty?
   - yes
     - are tx metadata available?
       - yes - send the message
       - no - write message to the topic
   - no - read messages from buffer topic till associated metadata are available and send them downstream

auditlog/log-enricher/src/main/java/io/debezium/demos/auditing/enricher/TopologyProducer.java
gunnarmorling: Yeah, that was a left-over from before where I had some logging code. Removed.

auditlog/log-enricher/src/main/java/io/debezium/demos/auditing/enricher/TopologyProducer.java
gunnarmorling: @jpechane, yes, that probably would work, too. Is it really worth the effort, though? That topic would have to be scoped to the specific KStreams instance, so it'd require some more thinking about the right partitioning strategy. The state store handles all this, plus it keeps the values in memory typically, so I'd assume it's better perf-wise, too. Keep in mind this buffer is usually empty or has a very small amount of entries.

So given I've already spent *way* more time on this than I had planned (and still need to write the blog post), I'd prefer to leave this as is for now. I can highlight the topic approach as an alternative implementation for the buffering in the blog post.

jpechane: @gunnarmorling Fine for me. I am just really surprised the is no OOTB implementation equivalent of `LinkedHashMap` as one would expect that processing in order is not an uncommon requirment. There is even `TimeOrderedKeyValueBuffer` but it is an internal only class.

hpgrahsl: @jpechane @gunnarmorling my 2 cents on this: I pretty much agree with Gunnar here, especially when having the blog post in mind for which this example primarily serves at the moment. However, in general I wouldn't recommend to make any assumptions on the ordering of keys for the underlying / backing state store. After all, the whole idea of the state store abstraction is to allow *any* implementation. E.g. InMemoryKeyValueStore uses a TreeMap internally without a comparator option which results in the natural ordering of keys. Since keys are represented with the Bytes class this means there will be lexicographic key ordering. Relying on all of this facts is IMHO highest-possible coupling of a solution towards the underlying concrete implementation which is why I personally would want to avoid it at all cost.

jpechane: @hpgrahsl Thanks! Just what do you think about my idea to use topic instead of keyvalue store? I am not advocating to change the PR, just I'd like to validate the viability of the approach.

auditlog/log-enricher/src/main/java/io/debezium/demos/auditing/enricher/TopologyProducer.java
gunnarmorling: Good idea. I'll give it a try (sorry for the late reply, lots of travel lately...)

auditlog/log-enricher/src/main/java/io/debezium/demos/auditing/enricher/TopologyProducer.java
```diff
@@ -0,0 +1,278 @@
...
+        StreamsBuilder builder = new StreamsBuilder();
+
+        StoreBuilder<KeyValueStore<Long, JsonObject>> streamBufferStateStore =
+                Stores
+                    .keyValueStoreBuilder(
+                        Stores.persistentKeyValueStore(STREAM_BUFFER_NAME),
+                        new Serdes.LongSerde(),
+                        new JsonObjectSerde()
+                    )
+                    .withCachingDisabled();
```
jpechane: @gunnarmorling Why is the caching disabled here?

auditlog/log-enricher/src/main/java/io/debezium/demos/auditing/enricher/TopologyProducer.java
```diff
@@ -0,0 +1,278 @@
...
+        builder.globalTable(txContextDataTopic, Materialized.as(STORE_NAME));
+
+        builder.<JsonObject, JsonObject>stream(vegetablesTopic)
+                // filter out any tombstones; for an audit log topic, time-based retention
+                // seems more reasonable than compaction
+                .filter((id, changeEvent) -> changeEvent != null)
+                // exclude snapshot events
+                .filter((id, changeEvent) -> !changeEvent.getString("op").equals("r"))
+                // enrich change events with transaction metadata via the statestore of the TX topic
+                .transform(() -> new ChangeEventEnricher(), STREAM_BUFFER_NAME)
```
jpechane: @gunnarmorling Just one technical note, you are using Transformer albeit you are not modifying the key. Using transforme can lead to repartitioning as it is expected that the key was modified. Processor does not have this problem.

auditlog/log-enricher/src/main/java/io/debezium/demos/auditing/enricher/TopologyProducer.java
gunnarmorling: @jpechane, it just seemed safer to me. Tbh. I'm not sure what the notion of "caching" exactly means here, so I decided to go for a better "safe than sorry" approach.

auditlog/log-enricher/src/main/java/io/debezium/demos/auditing/enricher/TopologyProducer.java
gunnarmorling: I think it is a valid approach, and I'll mention it as one possible alternative in the blog post. One potential "gotcha" may be that any unmanaged Kafka producer employed by the streaming pipeline code wouldn't be part of the Kafka Streams "exactly once" semantics, which may or may not cause trouble.

auditlog/log-enricher/src/main/java/io/debezium/demos/auditing/enricher/TopologyProducer.java
jpechane: @gunnarmorling Caching in this regard means that the changes are applied in memory and into a store backing topic. In case of crash the state is recovered from the state sotrage + changes from the backing topic are replayed. With cachin disabled every change is written directly into state store which can be I/O bottleneck.

auditlog/README.md
```diff
@@ -0,0 +1,107 @@
...
+## Deploy the Debezium Postgres Connector
+
+```console
+$ http PUT http://localhost:8083/connectors/inventory-connector/config < register-postgres.json
+```
+
+## Inserting Some Data and Observing the Audit Log
+
+```console
+$ http POST http://localhost:8080/vegetables 'Authorization: Bearer eyJraWQiOiJqd3Qua2V5IiwidHlwIjoiSldUIiwiYWxnIjoiUlMyNTYifQ.eyJzdWIiOiJmYXJtZXJib2IiLCJ1cG4iOiJmYXJtZXJib2IiLCJhdXRoX3RpbWUiOjE1NjY0NTgxMTMsImlzcyI6ImZhcm1zaG9wIiwiZ3JvdXBzIjpbImZhcm1lcnMiLCJjdXN0b21lcnMiXSwiZXhwIjo0MTAyNDQ0Nzk5LCJpYXQiOjE1NjY0NTgxMTMsImp0aSI6IjQyIn0.CscbJN8amqKryYvnVO1184J8F67HN2iTEjVN2VOPodcnoeOd7_iQVKUjC3h-ye5apkJjvAsQKrjzlrGCHRfl-n6jC9F7IkOtjoWnJ4wQ9BBo1SAtPw_Czt1I_Ujm-Kb1p5-BWACCBCVVFgYZTWP_laz5JZS7dIvs6VqoNnw7A4VpA6iPfTVfYlNY3u86-k1FvEg_hW-N9Y9RuihMsPuTdpHK5xdjCrJiD0VJ7-0eRQ8RXpycHuHN4xfmV8MqXBYjYSYDOhbnYbdQVbf0YJoFFqfb75my5olN-97ITsi2MS62W_y-RNT0qZrbytqINA3fF3VQsSY6VcaqRAeygrKm_Q' 'Date: Thu, 22 Aug 2019 08:12:31 GMT' name=Pear description=Yummy!
```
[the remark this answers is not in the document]
gunnarmorling: Hum, hum, I can't seem to reproduce this. Works fine with httpie 1.0.3. Can you upgrade and give it another try? Perhaps some platform issue? Where are you running on?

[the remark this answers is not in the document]
gunnarmorling: Also note this should be run all as one line, there shouldn't be any line breaks.

auditlog/log-enricher/src/main/java/io/debezium/demos/auditing/enricher/TopologyProducer.java
gunnarmorling: Done.

gunnarmorling:
Rebased and applied. Thanks a lot everyone for reviewing and all the feedback 👍  I think remaining issues can be sorted out incrementally. I'll begin to work on the blog post now.

auditlog/log-enricher/src/main/java/io/debezium/demos/auditing/enricher/TopologyProducer.java
hpgrahsl: @jpechane I think you are mixing up two different concepts, namely "caching vs. logging" here w.r.t state stores. both can be configured independently from one another. it's the logging that does the backing based on changelog topics which has in that sense nothing to do with the caching itself which may or may not be used.

PR #67, opened 2019-09-03
DBZ-1431

gunnarmorling:
https://issues.jboss.org/browse/DBZ-1431

@jpechane, as discussed.

### DIFF
--- a/kinesis/src/main/java/io/debezium/examples/kinesis/ChangeDataSender.java
+++ b/kinesis/src/main/java/io/debezium/examples/kinesis/ChangeDataSender.java
@@ -39,6 +39,7 @@ public class ChangeDataSender implements Runnable {
     private final Configuration config;
     private final JsonConverter valueConverter;
     private final AmazonKinesis kinesisClient;
+    private EmbeddedEngine engine;
 
     public ChangeDataSender() {
         config = Configuration.empty().withSystemProperties(Function.identity()).edit()
@@ -70,7 +71,7 @@ public class ChangeDataSender implements Runnable {
 
     @Override
     public void run() {
-        final EmbeddedEngine engine = EmbeddedEngine.create()
+        engine = EmbeddedEngine.create()
                 .using(config)
                 .using(this.getClass().getClassLoader())
                 .using(Clock.SYSTEM)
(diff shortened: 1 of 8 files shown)


kinesis/src/main/java/io/debezium/examples/kinesis/ChangeDataSender.java
```diff
@@ -85,20 +87,39 @@ public void run() {
...
         catch (InterruptedException e) {
-            Thread.interrupted();
+            Thread.currentThread().interrupt();
         }
     }
 
+    private void cleanUp() {
+        // Lingering MySQL thread without this...
+        try {
+            AbandonedConnectionCleanupThread.shutdown();
```
jpechane: Why is this needed? The thread in question is daemon thread. If the thread does not stop then there must be another thread lingering. Also the method in question is deprecated.

gunnarmorling: @jpechane, try to remove it, the thread will stick around. Happy to change it, if there's better ways to do it.

jpechane: @gunnarmorling I've just checked it for NATS - by the end of shutdown hook only daemon threads are alive (one of them is the MySQL dameon thread). hence I believe it is unnecessary. Otherwise you'd be required to excute the very same operation for any application that uses MySQL JDBC driver.

gunnarmorling:
Can you check with Kinesis, too (just comment out the actual handling logic
and have it raise an exception). Perhaps different driver versions are
involved and later ones don't expose it anymore, so updating the other
examples might help. For that one, without this call above, a thread named
"Abandoned connection cleanup thread" was sticking around.

jpechane:
@gunnarmorling Just checked the code of MySQL connector in Kinesis and the thread there is non-daemon. So I propose to either promote all three examples to `0.9.5.Final` or wait till `0.10.0.Final` is available and promote it to that one.

gunnarmorling:
@jpechane, reworked to use 0.9.5.Final consistently, removing the call for MySQL.

nats/README.md
```diff
@@ -17,7 +17,7 @@ This demo shows how to stream changes from MySQL database running on a local mac
 We will start a pre-populated MySQL database that is the same as used by the Debezium [tutorial](http://debezium.io/docs/tutorial/):
 
 ```
-docker run -it --rm --name mysql -p 3306:3306 -e MYSQL_ROOT_PASSWORD=debezium -e MYSQL_USER=mysqluser -e MYSQL_PASSWORD=mysqlpw debezium/example-mysql:0.9
+docker run -it --rm --name mysql -p 3306:3306 -e MYSQL_ROOT_PASSWORD=debezium -e MYSQL_USER=mysqluser -e MYSQL_PASSWORD=mysqlpw debezium/example-mysql:0.10
```
jpechane: @gunnarmorling Why `0.10` is used here and `0.9` in other cases?

gunnarmorling: This was an oversight, good catch. Updated to 0.9.

jpechane:
@gunnarmorling Applied, thanks

PR #68, opened 2019-09-11
DBZ-1474 Updated tutorial to show setting avro configuration in connector configuration

Naros:
https://issues.jboss.org/browse/DBZ-1474


Naros:
Hi @gunnarmorling.  This is the update to the tutorial we discussed earlier in the week after we learned that we could override and specify the Avro configuration at the connector level.  I decided to keep the worker approach and amend the tutorial to show how to accomplish the same at the connector level.  Let me know what you think.

PR #69, opened 2019-09-18
docs(outbox): fix typo

RafaelGSS:


jpechane:
@RafaelGSS Fixed one more link and applied, thanks for spotting!

PR #70, opened 2019-09-24
Updated PostgreSQL jdbc jar as well as kafka connect. New versions reflected in Dockerfile

davecramer:
There is a newer kafka-connect jar, but I'm not sure how much difference that makes if any.

I have mentioned to them to use rewriteInserts for batch insert in the pgjdbc jar but I doubt that has been implemented yet.


gunnarmorling:
Nice, thanks a lot, @davecramer!

PR #71, opened 2019-09-25
DBZ-1506 DBZ-1516 Updated several examples to support record deletion

Naros:
https://issues.jboss.org/browse/DBZ-1506
https://issues.jboss.org/browse/DBZ-1516


jpechane:
@Naros Hi, I've tried it and it does not work for me. I belive you need to take a look at `delete.handling.mode`/`drop.tombstones` setting of the sink SMT to make sure that the tombstone is passed to the sink connector.

gunnarmorling:
Good catch, the sink connector needs to be configured with drop.tombstones=true. 

@Naros, while at it, could you also change it to 

* use the new SMT class name
* merge the `connect` and `connect-json` services from the Compose file into one and adjust the README instructions accordingly; this still is based on the assumption that serializers cannot be configured per connector within a single Connect node

Could you also file a separate DBZ issue for replacing the WF Swarm consumer with one based on Quarkus.

Thanks!

gunnarmorling:
Another one to update would be this here: https://github.com/debezium/debezium-examples/tree/master/unwrap-smt

* Use current DBZ version
* Add DELETE example + config

Could also be done with a separate JIRA issue. Thanks!

Naros:
Pushed commit d9cf360 for Dave's benefit so he has a workable basis.  The support for deletes on the JDBC sink side required a minimum of kafka-connect-jdbc 5.3, so I upgraded it to the latest 5.3.1.  I'll finsh up the consolidation and push that later tonight.

gunnarmorling:
@Naros, could you also change it so that the JARs are obtained via wget instead of having them committed to the repo here. You can find precedent in other examples. The only reason I did it with the local files originally was to be able to build everything offline for the purposes of my demo. But that's not needed any longer, so there's no need for having these files checked in. Thanks!

Naros:
@gunnarmorling I believe this should be all set for one final review.  I also updated the MongoDB unwrap example as well and included a delete step too.

gunnarmorling:
Excellent, thanks a lot, @Naros!

@davecramer, if you check out the _unwrap-smt_ example now, it shows everything end to end, including deletes.

PR #72, opened 2019-10-07
[pgtoast] Initial import

gunnarmorling:
@jpechane, @Naros, the other day it came up on Twitter that people would like to learn how to deal with the "unchanged TOAST column" marker value. This example shows two approaches:

* using a trigger in the sink DB that ignores the value
* a KStreams app that replaces the value based on a statestore

Feedback very welcomed. I'll write a quick blog post, too (much shorter than the audit one).


postgres_toast/toast-value-store/src/main/java/io/debezium/demos/pgtoast/JsonObjectSerde.java
```diff
@@ -0,0 +1,91 @@
...
+import org.apache.kafka.common.serialization.Serde;
+import org.apache.kafka.common.serialization.Serializer;
+
+import io.quarkus.runtime.annotations.RegisterForReflection;
+
+/**
+ * A {@link Serde} that (de-)serializes JSON.
+ */
+@RegisterForReflection
+public class JsonObjectSerde implements Serde<JsonObject> {
```
jpechane: This is a cross-cutting concern that should be added into Debezium core.

postgres_toast/README.md
```diff
@@ -0,0 +1,130 @@
...
+This means that values larger than ca. 8 KB need to be stored using TOAST storage.
+This impacts replication messages coming from database,
+as the values that were stored using the TOAST mechanism and have not been changed are not included in the message,
+unless they are part of the table’s replica identity.
+
+The corresponding fields in Debezium change events will have configurable marker value in this case
+(defaulting to `__debezium_unavailable_value`).
+This demo shows two ways for handling such marker values:
+
+* ignoring any updates that'd set a column in a sink database to that marker value by means of a trigger installed n the sink database
```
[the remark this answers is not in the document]
gunnarmorling: Thanks, good catch. Fixed. Btw. how about using the "suggestion" feature for such changes? I.e. you'd just do fixes like that while reviewing, and I can then simply commit them. Makes it a bit easer.

postgres_toast/README.md
```diff
@@ -0,0 +1,130 @@
...
+    debezium/tooling:1.0 \
+    bash -c 'pgcli postgresql://postgresusersink:postgrespw@sink-db:5432/sinkdb'
+```
+
+```console
+sinkdb> select id, first_name, last_name, email, LEFT(biography,50) from inventorysink.customers where id = 1001;
+```
+
+This is done by means of a trigger on the `biography` column in the sink database table,
+which ignores any update to the special marker value (see _sink-db/schema-update.sql_).
```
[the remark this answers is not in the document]
gunnarmorling: Ok, I see. I'll reword.

postgres_toast/toast-value-store/src/main/java/io/debezium/demos/pgtoast/JsonObjectSerde.java
gunnarmorling: Yes, I was thinking about it, too. But than it's even more generic than Debezium. Unless we make it actually change-event aware and somehow expose "before" and "after" as dedicated objects for instance.

jpechane: I would not mind to make it Debezium specific so it is easy to consume.

postgres_toast/toast-value-store/src/main/java/io/debezium/demos/pgtoast/TopologyProducer.java
```diff
@@ -0,0 +1,47 @@
...
+        StreamsBuilder builder = new StreamsBuilder();
+
+        StoreBuilder<KeyValueStore<JsonObject, String>> streamBufferStateStore =
+                Stores
+                    .keyValueStoreBuilder(
+                        Stores.persistentKeyValueStore(INSTRUCTIONS_STORE),
+                        new JsonObjectSerde(),
+                        new Serdes.StringSerde()
+                    )
+                    .withCachingDisabled();
```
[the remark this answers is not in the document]
gunnarmorling: Ok, removed it.

postgres_toast/toast-value-store/src/main/java/io/debezium/demos/pgtoast/JsonObjectSerde.java
gunnarmorling: Logged as https://issues.jboss.org/browse/DBZ-1533

gunnarmorling:
Rebased and applied. Thanks for reviewing, @jpechane!

PR #73, opened 2019-10-14
audit log with kogito admin service

mswiderski:
it still uses quarkus master so it needs to be updated before final merge - expected version for it to work is 0.25.0


auditlog/admin-service/src/main/resources/application.properties
```diff
@@ -0,0 +1,24 @@
+# Configuration file
+# key = value
+
+quarkus.http.port=8085
```
gunnarmorling: For the sake of consistency with the other two services: can you remove this here, and instead change the exposed port of the service in the Compose file and override it in the README in the "Running locally" section?

auditlog/admin-service/src/main/java/io/debezium/demos/auditing/admin/TransactionEvent.java
```diff
@@ -0,0 +1,68 @@
+package io.debezium.demos.auditing.admin;
+
+import java.util.Date;
+
+import com.fasterxml.jackson.annotation.JsonFormat;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class TransactionEvent {
+    
+    @JsonProperty("before")
```
gunnarmorling: Tiny nit: I'd suggest to omit the annotation where the field name itself is fine.

auditlog/admin-service/pom.xml
```diff
@@ -0,0 +1,118 @@
...
+    <dependency>
+      <groupId>io.quarkus</groupId>
+      <artifactId>quarkus-junit5</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>io.rest-assured</groupId>
+      <artifactId>rest-assured</artifactId>
+      <scope>test</scope>
+    </dependency>
```
gunnarmorling: Are they needed actually?

auditlog/README.md
```diff
@@ -73,6 +73,49 @@ $ docker run -it --rm \
     -C -o beginning -q -u -t dbserver1.inventory.vegetable.enriched | jq ."
 ```
 
+## Administrate missing events
+
+There could be situations when there are missing transaction context data, for example
```
gunnarmorling: Probably worth pointing out here that this service is based on Kogito?

auditlog/admin-service/.gitignore
```diff
@@ -0,0 +1,35 @@
+# Eclipse
```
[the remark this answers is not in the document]
mswiderski: removed

auditlog/admin-service/src/main/resources/application.properties
```diff
@@ -0,0 +1,24 @@
...
+
+mp.messaging.incoming.transactions.connector=smallrye-kafka
+mp.messaging.incoming.transactions.bootstrap.servers=localhost:9092
```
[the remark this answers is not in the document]
mswiderski: I was thinking about it and was not sure how to name it... plus it could be slightly misleading for people using MP reactive messaging... so if possible I'd stick to these three props, wdyt?

auditlog/admin-service/src/main/resources/application.properties
mswiderski: removed and readme update

auditlog/admin-service/src/main/java/io/debezium/demos/auditing/admin/TransactionEvent.java
mswiderski: refactored as suggested

auditlog/admin-service/pom.xml
mswiderski: removed

auditlog/README.md
mswiderski: added few lines about kogito here

mswiderski:
@gunnarmorling I think I addressed all comments plus squashed the commits into one

auditlog/admin-service/src/main/resources/application.properties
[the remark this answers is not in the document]
gunnarmorling: Just to make sure we're on the same page: I was thinking about keeping the three props, but have them reference a single one:

bootstrap.servers=localhost:9092
mp.messaging.incoming.transactions.bootstrap.servers=${bootstrap.servers}
mp.messaging.incoming.vegetables.bootstrap.servers=${bootstrap.servers}
...

Would make it a bit simpler to override with a single value. If that's what you considered and decided against doing, that's cool, too. We can stick to the current approach then.

auditlog/README.md
gunnarmorling: Nice!

auditlog/admin-service/src/main/resources/application.properties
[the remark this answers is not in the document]
mswiderski: yes, that's what I considered... the thing was that I didn't want to use a property that might get in the way somehow... maybe it's a good thing to have it for MP messaging in general to allow to have properties on the top level to be inherited.

```
mp.messaging.bootstrap.servers=localhost:9092
```

so if you don't mind let's stick to current setup and I will open issue/question for Quarkus.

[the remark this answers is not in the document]
gunnarmorling: Gotcha. +1 for raising this with Quarkus folks (or mP reactive messaging). Thanks!

gunnarmorling:
Rebased and applied. Thanks a lot, @mswiderski!

PR #78, opened 2019-12-06
[tutorial] Externalization example

jpechane:


Naros:
Thanks for putting this together @jpechane, I think this is definitely going to be useful given the recent discussions about this feature.

gunnarmorling:
Nice! Also worth a quick blog post I'd say; the request comes up every so often.

jpechane:
@gunnarmorling https://github.com/debezium/debezium.github.io/pull/394

PR #79, opened 2019-12-11
[sql-server-read-replica] Read from read-only SQL Server replica

jpechane:


sql-server-read-replica/README.md
```diff
@@ -0,0 +1,81 @@
+# SQL Server Always On Cluster
+
+This demo automatically deploys the topology of services to stream from SQL Server read-only replica as described in the diagram.
+
+```
```
gunnarmorling: Nice chart :)

gunnarmorling:
Applied. Great work, thanks, @jpechane!

PR #81, opened 2019-12-20
DBZ-1478 Updates to Outbox Example

Naros:
https://issues.redhat.com/browse/DBZ-1478

This PR introduces the new order-service-quarkus module, a reimplementation of the Thorntail service order-service on the Quarkus platform.  This new module also makes use of the `debezium-quarkus-outbox` extension in PR https://github.com/debezium/debezium/pull/1184.

There are still the open action items for this PR that I'll be pushing to it:

- [x] Remove the non-Quarkus `shipment-service` module
- [x] Update version of Quarkus used by `shipment-service-quarkus` to `1.1.0.Final`
- [x] Provide a way to compose the build using either JVM or Native Images
There is already separate Dockerfile.jvm/Dockerfile.native for the `order-service-quarkus` module. 
 We need to replicate that to `shipment-service-quarkus` and then setup our compose files accordingly.
- [x] Remove the Thorntail based `order-service`.
- [ ] Optionally allow source application to write serialized Avro value to the payload and then use the pass-through converter to pass the byte array to Kafka.


Naros:
This is mostly for myself, but while migrating `shipment-service-quarkus` to `1.1.0.Final`, I'm hitting odd issues where I'm randomly seeing the service attempt to connect to the database using `root` despite being configured otherwise or the random `Unknown connector for orders` message.  I suspect this has to do with dependencies/configuration which likely matured since `0.11.0`, so a bit of investigation is needed.  I've opened a discussion with the Quarkus guys, hopefully I'll get some feedback in the coming days.

Naros:
So at this point all that remains is to address the optional Avro support and to make any adjustments that stem from the `debezium-quarkus` model being added to the main repo in the referenced PR above.

outbox/README.md
```diff
@@ -3,49 +3,75 @@
 This example demonstrates the "outbox pattern", an approach for letting services communicate in an asynchronous and reliable fashion.
 It accompanies [this post](https://debezium.io/blog/2019/02/19/reliable-microservices-data-exchange-with-the-outbox-pattern) on the Debezium blog.
 
-The sending service ("order-service") produces events in an "outbox" event table within its own local database.
+The sending service ("order-service-quarkus") produces events in an "outbox" event table within its own local database.
```
gunnarmorling: With the old services gone, let's just omit the "-quarkus" suffix in the names.

outbox/README.md
```diff
@@ -3,49 +3,75 @@
...
-which is functionally the same as the original one, but is implemented using the [Quarkus](https://quarkus.io) stack.
-This allows to build a native binary of that service, resulting in significantly less memory usage and faster start-up than the classic version (based on Thorntail).
+Both services are implemented using the [Quarkus](https://quarkus.io) stack.
+This allows building a native binary of each service, resulting in significantly less memory usage and faster start-up than the JVM-based version.
 
-## Execution
+## Building
 
-Prepare the Java components:
+Prepare the Java components by first performing a maven build.
```
gunnarmorling:
```suggestion
Prepare the Java components by first performing a Maven build.
```

outbox/README.md
```diff
@@ -3,49 +3,75 @@
...
-Start all components:
+This illustates the usage of Quarkus' `native` profile mode where the quarkus-maven-plugin will generate not only JVM-based artifacts but also native images.
+The `native` profile and `native-image.docker-build` environment variable can be omitted if native image artifacts are not wanted.
+
+## Environment
+
+Setup the necessary environment variables
 
 ```console
 $ export DEBEZIUM_VERSION=0.10
```
gunnarmorling: 1.0

outbox/README.md
```diff
@@ -3,49 +3,75 @@
...
 HTTP/1.1 201 Created
 ```
 
+## Call the various REST-based APIs
+
 Place a "create order" request with the order service:
 
 ```console
-$ http POST http://localhost:8080/order-service/rest/orders < resources/data/create-order-request.json
+$ http POST http://localhost:8080/rest/orders < resources/data/create-order-request.json
```
gunnarmorling: While at it, let's get rid of the "/rest" part, too.

outbox/README.md
```diff
@@ -74,7 +99,7 @@ Getting a session in the Postgres DB of the "order" service:
 $ docker run --tty --rm -i \
         --network outbox_default \
         debezium/tooling:1.0 \
-        bash -c 'pgcli postgresql://postgresuser:postgrespw@order-db:5432/orderdb'
+        bash -c 'pgcli postgresql://postgresuser:postgrespw@order-db-quarkus:5432/orderdb'
```
gunnarmorling: Same as above, let's remove "-quarkus" everywhere.

outbox/order-service-quarkus/src/main/java/io/debezium/examples/outbox/order/event/OrderLineUpdatedEvent.java
```diff
@@ -80,4 +60,9 @@ public JsonNode getPayload() {
     public String getType() {
         return "OrderLineUpdated";
     }
+
+    @Override
+    public Long getTimestamp() {
```
gunnarmorling: How about using `Instant` instead (applies more to the interface definition).

outbox/README.md
Naros: Do you want me to also change the directory names as well or simply the artifacts?

gunnarmorling: With only one flavour of these services I don't see name to encode the implementation choice into any names, i.e. let's simply use "order-service" everywhere.

outbox/order-service-quarkus/src/main/java/io/debezium/examples/outbox/order/event/OrderLineUpdatedEvent.java
Naros: Done, updated the extension PR and this in commit 7122617.

outbox/README.md
Naros: Removed

outbox/README.md
Naros: Applied

outbox/README.md
Naros: Suggestion applied

outbox/README.md
Naros: Renamed as requested.

outbox/README.md
```diff
@@ -8,44 +8,72 @@ Debezium captures the additions to this table and streams the events to consumer
...
+Prepare the Java components by first performing a Maven build.
 
 ```console
 $ mvn clean install -Pnative -Dnative-image.docker-build=true
 ```
 
+This illustates the usage of Quarkus' `native` profile mode where the quarkus-maven-plugin will generate not only JVM-based artifacts but also native images.
+The `native` profile and `native-image.docker-build` environment variable can be omitted if native image artifacts are not wanted.
+
+NOTE: There is some limitations with the Hibernate ORM Quarkus extension that prohibits running the Debezium Outbox Quarkus extension in native mode, so those arguments should be omitted for now.
```
gunnarmorling: I'd suggest to pull up this note right after line 12. Also is there an issue id we could link to?

outbox/docker-compose.yaml
```diff
@@ -65,50 +63,31 @@ services:
      - KAFKA_DEBUG=true
      - DEBUG_SUSPEND_FLAG=n
   order-service:
-    image: debezium-examples/outbox-order-service:latest
+    image: debezium-example/outbox-order-service:${DEBEZIUM_VERSION}
```
gunnarmorling: I tried to use `debezium-examples` consistently for all example images lately. So I'd prefer to stick to that (not all examples follow that pattern yet, some still use just `debezium`, which should get changed when touching such example).

outbox/shipment-service/src/main/java/io/debezium/examples/outbox/shipment/service/ShipmentService.java
```diff
@@ -18,6 +16,8 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.fasterxml.jackson.databind.JsonNode;
```
gunnarmorling: Fine for me, but I'm curious why you changed from Jsonb to Jackson?

outbox/shipment-service/src/main/java/io/debezium/examples/outbox/shipment/messagelog/ConsumedMessage.java
```diff
@@ -3,20 +3,18 @@
...
 
-import javax.persistence.Column;
 import javax.persistence.Entity;
 import javax.persistence.Id;
 
 @Entity
 public class ConsumedMessage {
 
     @Id
-    @Column(columnDefinition = "BINARY(16)")
```
gunnarmorling: This one created an error for my when Hibernate created the schema in Postgres. It was needed for MySQL/MariaDB, but Postgres supports UUID, so I could just remove it.

outbox/order-service/pom.xml
```diff
@@ -1,75 +1,129 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<project xmlns="http://maven.apache.org/POM/4.0.0"
-  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+<?xml version="1.0"?>
+<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns="http://maven.apache.org/POM/4.0.0"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
   <modelVersion>4.0.0</modelVersion>
-  <groupId>io.debezium.examples.outbox</groupId>
+  <groupId>com.github.examples.outbox</groupId>
```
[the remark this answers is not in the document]
Naros: Nope, entirely a mistake likely from merging PoCs - good catch!

outbox/README.md
Naros: I followed your advice, adding the two `ReflectiveClassBuildItem` entries we needed in our own extension got past the hurdle, so I've removed this in favor of having native support.

outbox/docker-compose.yaml
Naros: Fixed

outbox/shipment-service/src/main/java/io/debezium/examples/outbox/shipment/service/ShipmentService.java
Naros: Did I?  The shipment service was mostly left untouched as the Quarkus based shipment-service was already there and it looks like it was using Jackson already.  Since that seems to be the defacto in Quarkus, that's likely why nonetheless.

outbox/shipment-service/src/main/java/io/debezium/examples/outbox/shipment/messagelog/ConsumedMessage.java
Naros: But this does bring up an interesting question regarding the outbox table in the extension.  We store a UUID field there and I am wondering if we should consider allowing the user to define a custom column definition configuration option we plug-in here if supplied?

outbox/shipment-service/src/main/java/io/debezium/examples/outbox/shipment/service/ShipmentService.java
gunnarmorling: Ah, yes, that will be it. Thanks!

outbox/shipment-service/src/main/java/io/debezium/examples/outbox/shipment/messagelog/ConsumedMessage.java
gunnarmorling: So about that one, do you want to leave that for the follow-up? If so, could you update the issue [DBZ-1701](https://issues.redhat.com/browse/DBZ-1701) accordingly?

Naros: I think that'd probably best as I want to do some more testing across various data sources.  I actually logged the follow-ups in [DBZ-1711](https://issues.redhat.com/browse/DBZ-1711) since the other you referenced seemed to be related to CloudEvents.

Naros:
For clarity, I added the Avro serialization stuff to the [DBZ-1711](https://issues.redhat.com/browse/DBZ-1711) task as well so we can prepare to merge this PR as-is.

outbox/shipment-service/src/main/java/io/debezium/examples/outbox/shipment/messagelog/ConsumedMessage.java
gunnarmorling: Hah, yeah, was too late in the day for me when I wrote that comment. Thanks for logging a new one!

gunnarmorling:
Applied; thanks, @Naros!

Naros:
@gunnarmorling Pretty sure this example references `1.1.0-SNAPSHOT` so I'll add a follow-up as well to fix the version once 1.1 goes Final.

PR #82, opened 2019-12-21
Pgsql example

davecramer:
Enable example to work with postgres.

Need to change the name of the docker-compose.yaml file


kstreams-live-update/example-db/Dockerfile
```diff
@@ -1,3 +1,3 @@
-FROM debezium/example-mysql:0.10
```
gunnarmorling: Can we keep both? I.e. rename _example-db_ to _example-db-mysql_, and leave otherwise as is. Then add _example-db_postgres_ with your version. Then add a new service to the Compose file for PG, next to MySQL (alternatively, have two different Compose files). The Compose file would inject one of the DB names as env var into the event-source service.

kstreams-live-update/event-source/pom.xml
```diff
@@ -24,10 +24,10 @@
             <artifactId>hibernate-core</artifactId>
             <version>${version.org.hibernate}</version>
         </dependency>
-        <dependency>
-            <groupId>mysql</groupId>
-            <artifactId>mysql-connector-java</artifactId>
-            <version>8.0.12</version>
+	<dependency>
```
gunnarmorling: Should keep both as per my suggestion below.

gunnarmorling:
Thanks, @davecramer! Two comments inline, suggesting to keep both MySQL and PG, allowing to choose one when running docker-compose up.

davecramer:
@gunnarmorling whatever works for you. Having the second event-db dir works well I think

kstreams-live-update/docker-compose.yaml
```diff
@@ -50,11 +55,12 @@ services:
     build:
       context: event-source
     links:
-     - mysql
+     - postgres
     environment:
+     - DATABASE_SERVER=postgres
```
gunnarmorling: In  case you want to go for one Compose file with both services, you'd have to keep both links and both depends_on. Then you can inject the server name like so:

    - DATABASE_SERVER=${DATABASE_SERVER:-mysql}

Then, to run with Postgres do `DATABASE_SERVER=postgres docker-compose --build up`.
    environment:
      KAFKA_BOOTSTRAP_SERVERS: kafka:9092

davecramer: Honestly I think it is simpler to have 2 compose files. That way we don't have one service running that really isn't necessary

gunnarmorling:
👍 Ok, let's do that then.

gunnarmorling:
Hey @davecramer, Happy 2020! Just pushed one commit to this PR which brings back MySQL support as per the discussion above.

kstreams-live-update/docker-compose.yaml
gunnarmorling: Done.

kstreams-live-update/event-source/pom.xml
gunnarmorling: Done.

kstreams-live-update/example-db/Dockerfile
gunnarmorling: Done.

gunnarmorling:
@davecramer What's missing still is an update to the README.md and demo.md files. Specifically, for Postgres with pgoutput the steps for creating the publication are missing.

gunnarmorling:
@davecramer Ok, I've done the README update. But the subscription part is still missing. Could you add this one? Thanks!

davecramer:
@gunnarmorling so we don't need to fix the subscription part now so we should be able to apply this

gunnarmorling:
@davecramer, eventually circling back to this. So things are working now as per the updated README instructions? What made the subscription issue disappear?

davecramer:
@davecramer I wish it really was gone. I'm not convinced that it is. I am traveling at the moment and can't look into it. I'll try to update the README to deal with the situation where it doesn't start up correctly.

gunnarmorling:
Ok, thanks a lot; save travels!

>

gunnarmorling:
Pushed one more commit for some Dockerfile updates. Things work now AFAICS, so I'll rebase and merge. @davecramer, please let me know if you see any need for follow-up changes, or just send a PR for them :) Thanks a lot for this addition.

gunnarmorling:
Rebased and applied.

PR #90, opened 2020-01-26
increased version of debezium-quarkus-outbox

arjanvanderborg:
increased version of debezium-quarkus-outbox as snapshot version is not in maven repo


gunnarmorling:
Good catch, thanks!

PR #92, opened 2020-01-29
DBZ-1741 use https when accessing Maven repo

jpechane:
https://issues.redhat.com/browse/DBZ-1741


gunnarmorling:
Thanks, good catch. It makes me think that we probably should have some sort of CI on the examples repo, too. Not neccessarily running tests, but at least making sure the examples can be built, e.g. when bumping the version to the latest and greatest.

PR #93, opened 2020-02-03
fix kstream example aggregator script

c4tz:
Hey,

I just tried to start the `kstreams` example, but ran into the following error;

```
exec java -XX:+UseParallelGC -XX:GCTimeRatio=4 -XX:AdaptiveSizePolicyWeight=90 -XX:MinHeapFreeRatio=20 -XX:MaxHeapFreeRatio=40 -XX:+ExitOnOutOfMemoryError -cp .:/opt/poc-ddd-aggregates/lib/* io.debezium.examples.aggregation.StreamingAggregatesDDD dbserver1.inventory.customers dbserver1.inventory.addresses kafka:9092
usage: java -jar <package> io.debezium.examples.aggregation.StreamingAggregatesDDD <parent_topic> <children_topic> <bootstrap_servers>
```

I was able to fix it by using separate quotation for each parameter.


gunnarmorling:
Good catch, thanks a lot!

PR #100, opened 2020-03-20
fix env variables for postgres. POSTGRES_PASSWORD required for image to start

davecramer:
as per docs here https://github.com/docker-library/docs/tree/master/postgres

### DIFF
--- a/kstreams-live-update/docker-compose-postgres.yaml
+++ b/kstreams-live-update/docker-compose-postgres.yaml
@@ -30,9 +30,9 @@ services:
     environment:
      - PG_MODE=primary
      - PG_PRIMARY_PORT=5432
-     - PG_PRIMARY_USER=postgres
+     - POSTGRES_USER=postgres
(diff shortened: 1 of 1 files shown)


kstreams-live-update/docker-compose-postgres.yaml
```diff
@@ -30,9 +30,9 @@ services:
     environment:
      - PG_MODE=primary
      - PG_PRIMARY_PORT=5432
-     - PG_PRIMARY_USER=postgres
```
gunnarmorling: Will PORT above need updating, too?

davecramer: Seems to work fine, but probably because the port is the default port.

davecramer: I don't see any reference to the port in the docs

Naros:
This seems to be out-of-date, closing.  Thanks @davecramer!.

PR #102, opened 2020-04-01
[monitoring] Show all connectors in dashboard

jpechane:


jpechane:
@gunnarmorling Just a small improvement

gunnarmorling:
Cool; btw. do we have a blog post on this one yet?

jpechane:
@gunnarmorling I must admit that I don't remeber it but I think we don't

gunnarmorling:
Could be a nice topic then :) But lesser in terms of prio than some others
we've already planned and discussed.

>

PR #110, opened 2020-04-26
Updated to use camel-coap-kafka-connector 0.1.0

valdar:


valdar:
Since the first version of `camel-kafka-connect` has been released, this is an update use it.

gunnarmorling:
Nice, thanks a lot! With the Camel COAP connector now available as a ZIP, we don't even need that little helper Maven module at all any longer (it only existed for being able to use the Maven dependency plug-in). We could simply wget + unpack the archive in the Dockerfile, as e.g. here in another example: https://github.com/debezium/debezium-examples/blob/master/unwrap-smt/debezium-jdbc-es/Dockerfile. Could you update the PR accordingly perhaps?

valdar:
Sure!

gunnarmorling:
Excellent, much appreciate!

valdar:
@gunnarmorling it should be done.

gunnarmorling:
Rebased and applied. Thanks a lot, @valdar!

PR #113, opened 2020-05-12
[kinesis] Updating to current Debezium version

gunnarmorling:
@jpechane, as discussed.


jpechane:
@gunnarmorling Thanks!

PR #114, opened 2020-05-15
[auditlog] Upgrading to current version of Debezium and Quarkus;

gunnarmorling:
* Kafka Streams 2.3 brought an incompatability around state stores, hence the change to TimestampedKeyValueStore (see https://kafka.apache.org/23/documentation/streams/upgrade-guide#streams_api_changes_230)
* The Kogito admin service needed updating to ignore the new "transaction" field and use latest MP Reactive Messaging APIs

@Naros, could you check out this one? Note there's still one issue with the admin service, it currently fails to send the TX metadata events to Kafka. I'm discussing with @cescoffier why that is.

@cescoffier, here's the stacktrace:

```
admin-service_1       | Caused by: java.lang.ClassCastException: io.smallrye.reactive.messaging.kafka.OutgoingKafkaRecord cannot be cast to java.lang.String
admin-service_1       | 	at org.apache.kafka.common.serialization.StringSerializer.serialize(StringSerializer.java:28)
admin-service_1       | 	at org.apache.kafka.common.serialization.Serializer.serialize(Serializer.java:62)
admin-service_1       | 	at org.apache.kafka.clients.producer.KafkaProducer.doSend(KafkaProducer.java:903)
admin-service_1       | 	at org.apache.kafka.clients.producer.KafkaProducer.send(KafkaProducer.java:865)
admin-service_1       | 	at io.vertx.kafka.client.producer.impl.KafkaWriteStreamImpl.lambda$send$4(KafkaWriteStreamImpl.java:93)
admin-service_1       | 	at io.vertx.core.impl.ContextImpl.lambda$executeBlocking$2(ContextImpl.java:316)
admin-service_1       | 	at io.vertx.core.impl.TaskQueue.run(TaskQueue.java:76)
admin-service_1       | 	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
admin-service_1       | 	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
admin-service_1       | 	at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
admin-service_1       | 	at java.lang.Thread.run(Thread.java:748)
```

The outgoing channel is defined as

```
mp.messaging.outgoing.missingtransactions.topic=dbserver1.inventory.transaction_context_data
mp.messaging.outgoing.missingtransactions.key.serializer=org.apache.kafka.common.serialization.StringSerializer
mp.messaging.outgoing.missingtransactions.value.serializer=org.apache.kafka.common.serialization.StringSerializer
```

It's used via `@Channel("missingtransactions") Emitter<KafkaRecord<String, String>> emitter;` in the application. It's unclear to me why the `OutgoingKafkaRecord` itself is passed to the serializer instead of the key/value.

### DIFF
--- a/auditlog/admin-service/src/main/java/io/debezium/demos/auditing/admin/wih/SendTransactionEventWIHandler.java
+++ b/auditlog/admin-service/src/main/java/io/debezium/demos/auditing/admin/wih/SendTransactionEventWIHandler.java
@@ -4,6 +4,8 @@ import java.util.Collections;
 
 import javax.enterprise.context.ApplicationScoped;
 
+import org.eclipse.microprofile.reactive.messaging.Channel;
+import org.eclipse.microprofile.reactive.messaging.Emitter;
 import org.kie.api.runtime.process.WorkItem;
 import org.kie.api.runtime.process.WorkItemHandler;
 import org.kie.api.runtime.process.WorkItemManager;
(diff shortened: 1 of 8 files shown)


auditlog/admin-service/src/main/java/io/debezium/demos/auditing/admin/wih/SendTransactionEventWIHandler.java
```diff
@@ -14,43 +16,41 @@
...
 @ApplicationScoped
 public class SendTransactionEventWIHandler implements WorkItemHandler {
 
     private ObjectMapper json = new ObjectMapper();
-    
+
     @Channel("missingtransactions")
-    Emitter<KafkaMessage<String, String>> emitter;
-    
+    Emitter<KafkaRecord<String, String>> emitter;
```
cescoffier, line 27: So you cannot do this anymore :-(

Use:

```
Emitter<String> emitter;
```

It will let you send the KafkaRecord or a payload directly.

gunnarmorling: Ok, yeah, but how can we specific key and value?

cescoffier: Yes, you can continue sending your KafkaRecord as you did.

gunnarmorling: I.e. how do I send `KafkaRecord` via `Emitter <String>`?

auditlog/admin-service/src/main/java/io/debezium/demos/auditing/admin/wih/SendTransactionEventWIHandler.java
```diff
@@ -14,43 +16,41 @@
...
             event.setAfter(data);
-    
+
             String key = json.writeValueAsString(Collections.singletonMap("transaction_id", vegetable.getSource().getTransactionId()));
             String value = json.writeValueAsString(event);
-            
-            emitter.send(KafkaMessage.of(key, value));
-            
+
+            emitter.send(KafkaRecord.of(key, value));
```
cescoffier, line 52: This does not change.

auditlog/admin-service/src/main/java/io/debezium/demos/auditing/admin/wih/SendTransactionEventWIHandler.java
cescoffier: Like this:

```
 emitter.send(KafkaRecord.of(key, value));
```

The emitter parameter is the payload type. So you can send the payload directly, or a `Message<X>` like a `KafkaRecord<String, X>`

gunnarmorling: Ok, gotcha. Yes, that works. FWIW, I found the earlier API more comprehensible, for Kafka with its notion of key and value it's quite confusing to have an emitter interface with only one type parameter.

In any case, thanks a lot for the quick reply. Will update this PR now.

cescoffier:
I totally agree with you. I will introduce a specific MessageEmitter soon.

gunnarmorling:
@Naros, this one is complete from my side.

Naros:
Thanks @gunnarmorling.

PR #115, opened 2020-06-15
DBZ-2163 Topic name mapper demo

jpechane:
https://issues.redhat.com/browse/DBZ-2163

### DIFF
--- a/debezium-server-name-mapper/README.md
+++ b/debezium-server-name-mapper/README.md
@@ -1,14 +1,14 @@
 # Custom Topic Name Mapper
 
-This example demonstrates how to develop and deploy a bean that would allow the developer to implement custom topic naming policy.
+This example demonstrates how to develop and deploy a bean that would allow the developer to implement a custom topic naming policy.
 
 The implementation of the custom naming policy is in the class `PrefixingNameMapper`.
-The class is marked as `@Dependent` is it is injected into the using sink class.
-It exposes `mapper.prefix` configuration option that defines the sting that is inserted before the original topic name.
+The class is marked as `@Dependent` and it is injected into the configured sink adaptor class.
+It exposes a configuration option named `mapper.prefix` which defines a string that is inserted before the original topic name.
 
-The configuration option is configured either via `application.properties` configuration file or via other ways defined in Microprofile Config specification.
+The configuration option is configured either via `application.properties` configuration file or via other ways defined in the Microprofile Config specification, e.g. an environment variable.
 
-As an example - a message intended to be delivered to topic `server.table` would be routed to topic `superprefix.server.table`.
+As an example, when setting `mapper.prefix` to the value `superprefix`, then a message intended to be delivered to the topic `server.schema.table` would be routed to topic `superprefix.server.schema.table`.
 
 
 ## Topology
--- a/debezium-server-name-mapper/README.md
+++ b/debezium-server-name-mapper/README.md
@@ -1,0 +1,77 @@
+# Custom Topic Name Mapper
+
+This example demonstrates how to develop and deploy a bean that would allow the developer to implement custom topic naming policy.
+
+The implementation of the custom naming policy is in the class `PrefixingNameMapper`.
+The class is marked as `@Dependent` is it is injected into the using sink class.
+It exposes `mapper.prefix` configuration option that defines the sting that is inserted before the original topic name.
+
+The configuration option is configured either via `application.properties` configuration file or via other ways defined in Microprofile Config specification.
+
+As an example - a message intended to be delivered to topic `server.table` would be routed to topic `superprefix.server.table`.
+
+
+## Topology
+
+This demo uses Apache Pulsar as the sink and the standard Debezium example PostgreSQL database.
+Both Apache Pulsar and the source database are deployed via Docker Compose file.
+
+
+## How to run
+
+From terminal start the source database and the sink system:
+
+```
+$ docker-compose up
+```
+
+In another terminal build the custom naming policy class and the runner JAR to start the application:
+
+```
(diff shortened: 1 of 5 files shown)


debezium-server-name-mapper/src/main/resources/application.properties
```diff
@@ -0,0 +1,18 @@
...
+debezium.source.offset.flush.interval.ms=0
+debezium.source.database.hostname=localhost
+debezium.source.database.port=5432
+debezium.source.database.user=postgres
+debezium.source.database.password=postgres
+debezium.source.database.dbname=postgres
+debezium.source.database.server.name=tutorial
+debezium.source.schema.whitelist=inventory
+
+# Needed till https://issues.redhat.com/browse/DBZ-2192 is released
```
gunnarmorling, line 14: That's merged now (also answers my question on the issue :-). Can you remove this and rebase?

debezium-server-name-mapper/pom.xml
```diff
@@ -0,0 +1,69 @@
...
+    <artifactId>debezium-server-name-mapper</artifactId>
+    <version>1.0.0-SNAPSHOT</version>
+
+    <properties>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <maven.compiler.source>1.8</maven.compiler.source>
+        <maven.compiler.target>1.8</maven.compiler.target>
+
+        <database.hostname>localhost</database.hostname>
+        <database.port>3306</database.port>
```
gunnarmorling, line 14: Looks odd given we use Postgres? Are these props needed at all?

debezium-server-name-mapper/README.md
```diff
@@ -0,0 +1,77 @@
+# Custom Topic Name Mapper
+
+This example demonstrates how to develop and deploy a bean that would allow the developer to implement custom topic naming policy.
```
gunnarmorling, line 3:
```suggestion
This example demonstrates how to develop and deploy a bean that would allow the developer to implement a custom topic naming policy.
```

debezium-server-name-mapper/README.md
```diff
@@ -0,0 +1,77 @@
...
+
+The implementation of the custom naming policy is in the class `PrefixingNameMapper`.
+The class is marked as `@Dependent` is it is injected into the using sink class.
```
gunnarmorling, line 6:
```suggestion
The class is marked as `@Dependent` and it is injected into the configured sink adaptor class.
```

debezium-server-name-mapper/README.md
```diff
@@ -0,0 +1,77 @@
...
+It exposes `mapper.prefix` configuration option that defines the sting that is inserted before the original topic name.
```
gunnarmorling, line 7:
```suggestion
It exposes a configuration option named `mapper.prefix` which defines a string that is inserted before the original topic name.
```

debezium-server-name-mapper/README.md
```diff
@@ -0,0 +1,77 @@
...
+
+The configuration option is configured either via `application.properties` configuration file or via other ways defined in Microprofile Config specification.
```
gunnarmorling, line 9:
```suggestion
The configuration option is configured either via `application.properties` configuration file or via other ways defined in the Microprofile Config specification, e.g. an environment variable.
```

debezium-server-name-mapper/README.md
```diff
@@ -0,0 +1,77 @@
...
+
+As an example - a message intended to be delivered to topic `server.table` would be routed to topic `superprefix.server.table`.
```
gunnarmorling, line 11:
```suggestion
As an example, when setting `mapper.prefix` to the value `superprefix`, then a message intended to be delivered to the topic `server.schema.table` would be routed to topic `superprefix.server.schema.table`.
```

debezium-server-name-mapper/pom.xml
```diff
@@ -0,0 +1,69 @@
...
+
+        <mysql.init.timeout>30000</mysql.init.timeout>
+
+        <version.debezium>1.2.0-SNAPSHOT</version.debezium>
+        <version.debezium.tag>1.2</version.debezium.tag>
+        <version.quarkus>1.5.0.Final</version.quarkus>
+    </properties>
+
+    <dependencyManagement>
+        <dependencies>
```
gunnarmorling, line 30: Instead of using Debezium Server as a dependency, could we rather build this as a simple JAR (i.e. no Quarkus app itself) and then show how the built JAR is added to the lib folder of the Quarkus application? Would this not be a more common usage?

debezium-server-name-mapper/README.md
gunnarmorling: Should also mention that this is specific to Debezium Server, including a link to its docs.

debezium-server-name-mapper/src/main/resources/application.properties
jpechane: No, I can't. Will remove only after the JARs with fix are released.

debezium-server-name-mapper/pom.xml
jpechane: Good catch, it is copy/paste error

debezium-server-name-mapper/pom.xml
jpechane: @gunnarmorling IMHO we can't. Is not it that way that DI in Quarkus is resolved in build time not in runtime?

gunnarmorling: Yes, you're right. That was a bit of a brain fart on my end :)

gunnarmorling:
Nice one; applied.

PR #123, opened 2020-07-08
Update application.properties

limadelrey:
Minor changes on available configurations in order to include "payload" field options


limadelrey:
Just a little detail, eager to contribute with something more substantial in the future!

Thank you, I really appreciate it. I admire the work you put into Debezium. Hopefully more people will be encouraged to use it after reading that blog post 👍

PR #124, opened 2020-07-10
DBZ-2083 fix tutorial for Apicurio converter

rk3rn3r:
After debezium/docker-images#173 and debezium/docker-images#175 this should update the docs to use new  `ENABLE_APICURIO_CONVERTERS` env var to enable Apicurio converters.


gunnarmorling:
Do we say to use 1.2 in the tutorial README already?

rk3rn3r:
@gunnarmorling oh damn. but it's easy to upgrade. should I?

gunnarmorling:
@rk3rn3r, upgrading all the examples would also mean running through their instructions and making sure they still work as described. Definitely independent from that issue here. But it would be nice to do it, why not. So if you're up for it, please log a Jira issue and feel free to grab it. Could be a nice exercise actually to run through all of them and get to know them. But I'd suggest to time-cap it, i.e. not spending a week doing so.

PR #135, opened 2020-10-23
DBZ-2693 Added entry-point script for OCP environment

jcechace:
Adds entry-point script for OCP with the ability to execute scripts prior to any other db setup. 
Unforuntale DB2 in OCP seems to have issues to start when database directory in emptyDIr volume is not empty and for some reason that happens sometimes.

### DIFF
--- a/tutorial/debezium-db2-init/db2server/Dockerfile
+++ b/tutorial/debezium-db2-init/db2server/Dockerfile
@@ -19,3 +19,5 @@ RUN mkdir /var/custom
 RUN chmod -R  777 /var/custom
 
 ADD cdcsetup.sh /var/custom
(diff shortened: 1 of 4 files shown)


tutorial/debezium-db2-init/db2server/setup_db2_instance.sh
```diff
@@ -0,0 +1,210 @@
+#!/bin/bash
```
jpechane, line 1: @jcechace Is this file really needed? I don't see it used anywhere

jcechace: You are correct. That is the orignal from ibm image

PR #137, opened 2020-10-26
Update debezium-dashboard.json

Kiegen:
The below changes is quite useful when running debezium in distributed mode and you have multiple instances persisting to a single prometheus.
Added job variable
Removed context variable and harcoded to specific row.
Updated regex on variables to use value's for selected connector.
Created row for jvm metrics

### DIFF
--- a/monitoring/debezium-grafana/debezium-dashboard.json
+++ b/monitoring/debezium-grafana/debezium-dashboard.json
@@ -12,210 +12,56 @@
       }
     ]
   },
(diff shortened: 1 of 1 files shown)


jpechane:
@Kiegen Hi, thanks for the PR. It seems to me that it is tailored to MySQL but the original dashboard was connector agnostic. Would it be possible to introduce only the advertised fetaures but keep the genericity of the connector metrics?

PR #138, opened 2020-12-03
DBZ-1818 Extend outbox example with tracing

jpechane:
Requires artifacts from https://github.com/debezium/debezium/pull/1989

Strimzi image is used instead of Debezium's due to baked support for OpenTracing. 
This PR needs to stay as draft till the main PR is not merged and at least snapshots are published.
As workaround the plugin path is exposed as volume and built postgres connector must be placed in it.

Follow standard docs to start the demo.

Jaeger UI is available at http://localhost:16686/ 

A sample tracing tree is https://pasteboard.co/JDdk40a.png


outbox/docker-compose.yaml
```diff
@@ -55,23 +64,39 @@ services:
...
+         consumer.interceptor.classes=io.opentracing.contrib.kafka.TracingConsumerInterceptor
+         producer.interceptor.classes=io.opentracing.contrib.kafka.TracingProducerInterceptor
+     - JAEGER_SERVICE_NAME=kafka-connect
+     - JAEGER_AGENT_HOST=jaeger
+     - STRIMZI_TRACING=jaeger
      - KAFKA_DEBUG=true
      - DEBUG_SUSPEND_FLAG=n
+     - JAVA_DEBUG_PORT=*:5005
+    volumes:
+      - $HOME/tmp/debezium:/opt/kafka/plugins
```
gunnarmorling, line 98: Can we find a more portable way for this one :)

jpechane: See the comment, this is workaround till at least snapshot with the function is out. Then the curl will be used to build custom image. Till this is done the PR will stay as draft

gunnarmorling:
@jpechane, building with Java 1.8 yields a warning by Quarkus, i.e. we should upgrade the container images to use JDK 11 instead. Would you like to do this as part of this PR?

gunnarmorling:
Another housekeeping comment: can we remove the "outbox" prefix from the service names? It's kinda confusing when looking at the spans in the Jaeger UI.

jpechane:
@gunnarmorling IIUC I am building the services locally not in docker container so upgrading the container images will not the warning go away as this is built with your local JDK
regarding the `prefix` I must admit I am a little bit confused as the service name is properly set at docker-compose level yet the Maven module name is used. I am trying to investigate wy is that

gunnarmorling:
> @gunnarmorling IIUC I am building the services locally not in docker container so upgrading the container images will not the warning go away as this is built with your local JDK

The container image only has Java 8, i.e. you need to build for that version on your host. This is where the Quarkus build raises a warning (at least for me). If we want to build for 11, we need to update the container image in order to run the app.

outbox/docker-compose.yaml
```diff
@@ -1,5 +1,14 @@
 version: '2.1'
 services:
+  jaeger:
+    image: jaegertracing/all-in-one:1
+    ports:
+     - 5775:5775/udp
+     - 6831:6831/udp
+     - 6832:6832/udp
```
jpkrohling, line 8: In typical cases for Java applications, you should only need the ports `6831/udp` (agent) and `16686/tcp` (UI). You'll need the port `14268/tcp` only if you are exporting data directly from the client (your application) to the collector, bypassing the agent. In that case, you don't need `6831/udp`.

gunnarmorling:
@jpechane, core PR is merged. This should be good to go, too, once we can pull in the updated core artifacts.

gunnarmorling:
@jpechane, we didn't talk about tracing on the broker yet; can we enable a span there, too?

jpechane:
@gunnarmorling I've not found any way how to support broker side tracing and even Strimzi docs declares it is not available.

gunnarmorling:
Gasp, out of anything, I thought that'd be the one thing Strimzi would support. Never mind then.

jpechane:
@gunnarmorling The example should now be compelte. Connector is deployed from Maven central and a short notice into docs together with a sample screenshot was added.

outbox/docker-compose.yaml
gunnarmorling: @jpechane, can we limit the number of ports?

jpechane: @gunnarmorling I think it is unnecessary. It mimics official jaeger docs https://www.jaegertracing.io/docs/1.21/getting-started/#all-in-one

outbox/docker-compose.yaml
```diff
@@ -55,23 +64,38 @@ services:
     ports:
       - "${PGADMIN_PORT:-5050}:80"
   connect:
-    image: debezium/connect:${DEBEZIUM_VERSION}
+    image: debezium/strimzi-connect
+    build:
+      context: debezium-strimzi
```
gunnarmorling, line 69: This directory seems to be missing?

gunnarmorling:
@jpechane, pushed a commit for some clean-up. Two comments above: strimzi context dir is missing, and whether we can limit the number of exposed ports for Jaeger.

jpechane:
@gunnarmorling Done

gunnarmorling (approved):

gunnarmorling:
Woohoo, great stuff!

PR #139, opened 2020-12-04
DBZ-1463 Setup Cassandra Tutorial

ani-sha:
https://issues.redhat.com/browse/DBZ-1463

Creating a Draft PR for review, need to add some test data and setup README. 
cc: @jpechane


tutorial/README.md
```diff
@@ -341,6 +342,28 @@ db2 "INSERT INTO DB2INST1.CUSTOMERS(first_name, last_name, email) VALUES ('John'
...
+
+# Consume messages from a Debezium topic
+docker-compose -f docker-compose-cassandra.yaml exec kafka /kafka/bin/kafka-console-consumer.sh \
+    --bootstrap-server kafka:9092 \
+    --from-beginning \
+    --property print.key=true \
+    --topic server1.dbo.customers
+
+# Initialize and insert some test data 
+docker-compose -f docker-compose-cassandra.yaml exec cassandra bash -c 'cqlsh --file="/debezium/inventory.cql"'
```
jpechane, line 361: This can be done instratup script automatically, right after the server is started and initialized. No need for user to do it manually.

tutorial/debezium-cassandra-init/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+    CASSANDRA_YAML=/opt/cassandra/conf \
+    DEBEZIUM_HOME=/debezium
+
+COPY cassandra.yaml $CASSANDRA_YAML
+
+RUN mkdir $DEBEZIUM_HOME
+RUN curl -fSL -o $DEBEZIUM_HOME/debezium-connector-cassandra.jar \
+                 $MAVEN_CENTRAL/io/debezium/debezium-connector-cassandra/$DEBEZIUM_VERSION/debezium-connector-cassandra-$DEBEZIUM_VERSION-jar-with-dependencies.jar
+
+COPY log4j.properties $DEBEZIUM_HOME
```
jpechane, line 19: I've just checked the Docker file docs and it seems that the `COPY` instuctions can be merged into a single one as the destination dir is same for all files

tutorial/debezium-cassandra-init/log4j.properties
```diff
@@ -0,0 +1,6 @@
+log4j.rootLogger=DEBUG, A1
```
jpechane, line 1: Switch to `INFO`, `DEBUG` is probably unnecessary for demo purposes

tutorial/README.md
```diff
@@ -341,6 +342,28 @@ db2 "INSERT INTO DB2INST1.CUSTOMERS(first_name, last_name, email) VALUES ('John'
...
+
+# Shut down the cluster
```
jpechane, line 363: Please see https://github.com/debezium/debezium-examples/tree/master/tutorial#using-postgres and add an equivalent of step `Modify records in the database via Postgres client`

jpechane (changes requested):
@ani-sha Overall good work! I've just left a couple of minor comments

ani-sha:
@jpechane Added the change suggestions!

tutorial/debezium-cassandra-init/Dockerfile
```diff
@@ -0,0 +1,28 @@
+FROM cassandra:3.11
+
+RUN apt-get update && \
+    apt-get install -y wget curl sudo vim nano git gnupg
+
+RUN sudo apt install -y openjdk-8-jdk
```
jpechane, line 6: Could be merged with `RUN` above

tutorial/debezium-cassandra-init/Dockerfile
```diff
@@ -0,0 +1,28 @@
+FROM cassandra:3.11
+
+RUN apt-get update && \
+    apt-get install -y wget curl sudo vim nano git gnupg
```
jpechane, line 4: Are all the utilities but `wget` necessary?

ani-sha: Removing the unused ones, we don't need them now.

tutorial/debezium-cassandra-init/Dockerfile
```diff
@@ -0,0 +1,27 @@
+FROM cassandra:3.11
+
+RUN apt-get update && \
+    apt-get install -y curl sudo  && \
+    sudo apt install -y openjdk-8-jdk
+
+ENV DEBEZIUM_VERSION=1.4.0.Alpha2 \
```
gunnarmorling, line 7: Let's use 1.3.1.Final then, so to be in sync with 1.3 above.

ani-sha: yes @gunnarmorling! Updated.

tutorial/debezium-cassandra-init/cassandra.yaml
```diff
@@ -0,0 +1,1021 @@
+# Cassandra storage config YAML
```
gunnarmorling, line 1: Instead of copying the entire Cassandra configuration file, could we also take the default one and just replace/amend the values we actually need to customize, e.g. using `sed`? I'm concerned about maintenance effort of this, e.g. when it comes to upgrading to a newer version. Right now it's hard to see what configuration changes are needed in this file.

tutorial/debezium-cassandra-init/log4j.properties
```diff
@@ -0,0 +1,6 @@
+log4j.rootLogger=INFO, A1
+
+log4j.appender.A1=org.apache.log4j.ConsoleAppender
+
+log4j.appender.A1.layout=org.apache.log4j.PatternLayout
+log4j.appender.A1.layout.ConversionPattern=%-4r [%t] %-5p %c %x - %m%n
```
gunnarmorling, line 6: There are some new-lines at EOF missing in this PR, here and elsewhere.

tutorial/README.md
gunnarmorling: This still seems to be missing? I.e. the actual CQL statements within cqlsh for inserting/updating some records, akin to what we do for PG and the others?

gunnarmorling (commented):
Hey @ani-sha, some comments inline. Let me know if anything is not quite clear. Thanks a lot!

tutorial/debezium-cassandra-init/cassandra.yaml
ani-sha: There were particularly 8-9 modifications that were needed to the original config file. This was @jpechane's idea of putting our customized configuration file. I don't think it will be trouble in upgrading to newer versions. @jpechane any points to add?

tutorial/debezium-cassandra-init/log4j.properties
ani-sha: No, I don't think any lines are missing. This file contains this much only.

tutorial/README.md
ani-sha: Should I add some CQL statements (Insert, Update, etc) in the README?  
I have already added the step for ` Modify records in the database via Cassandra client` where the user can do any modifications of their choice within the cqlsh under keyspace `testdb`.

tutorial/debezium-cassandra-init/cassandra.yaml
jpechane: @gunnarmorling Yes, that was my recommendation. As Cassandra is mostly developed in 4.x tree I don't suppose any important changes in 3.x default config file

tutorial/debezium-cassandra-init/log4j.properties
jpechane: @ani-sha Gunnar means the line end (LF char) at the last line of the file. Not that any text is missing

tutorial/README.md
jpechane: @gunnarmorling We provide it only few of dbs like MongoDB. Definitely not in postgres I pointed to @ani-sha like an example.

gunnarmorling: > Should I add some CQL statements (Insert, Update, etc) in the README?

Yes, exactly. I thought we'd consistently do that for the others, but turns out we don't :)

tutorial/debezium-cassandra-init/cassandra.yaml
gunnarmorling: > As Cassandra is mostly developed in 4.x tree I don't suppose any important changes in 3.x default config file

That's my point though. When we update to 4.x at some point in the future, we'll likely bring in its default configuration file, just to make sure we don't miss any values they may have newly introduced. But then we need to re-apply our own changes, and it's hard to find out which those were. You'll have to go back to a default 3.x file and compare what we've changed.

If you don't warm up for the sed approach, then let's at least split up the commit into two:

* The first which adds the unmodified configuration file and everything else
* The second which does the required changes to the configuration file

That way we'll easily know what has to be adjusted whenever we need to update that file.

ahmedjami:
Hi,

Thanks a lot for this good work, it will be very useful for everyone who would like to start using Cassandra-connector : )

A few minor points:
With commitlog_segment_size_in_mb = 32M (Default value) and what we insert into product table, I'm not sure that we will have a full commitLog Segment that it will be transferred by Cassandra (when it's flushed) into cdc_raw and be consumed by the connector.

To have more data/traffic, we can use cassandra-stress (installed with Cassandra) to populate more data as described [here](https://github.com/ahmedjami/debezium-cassandra) OR change the default value of commitlog_segment_size_in_mb : )

tutorial/debezium-cassandra-init/cassandra.yaml
jpechane: @gunnarmorling Or even better consolidate all changes at the and of file and mark them as scuh

ahmedjami:
Also, once this PR merged, I can add a "Getting Started" section in cassandra-connector repo that redirects to the tutorial example.

tutorial/debezium-cassandra-init/cassandra.yaml
ani-sha: @gunnarmorling @jpechane I think we need to use `sed`. Because we have to enable as well as disable properties from the configuration file. I just checked these are these modifications are to be done.
```
1. Remove enable_sasi_indexes
2. Remove replica_filtering_protection (cached_rows_warn_threshold: 2000 cached_rows_fail_threshold: 32000)
3. Add commitlog_directory
4. Add hints_directory
5. Add cdc_raw_directory
6. Add saved_caches_directory
7. Add data_file_directories 
8. Change enable_cdc to true
9. Change IPs from 172.18.0.4 to 172.0.0.1 in multiple places.
```

gunnarmorling: @ani-sha, ok, but the "separate commit" strategy  should work, too, right? I don't mind too much which way tbh. Just assume you'd have to look into this in 6 months from now: how can you prepare future Anisha to know what needs to change at that point :)

gunnarmorling:
Hey @ahmedjami, thanks for chiming in!

> Thanks a lot for this good work, it will be very useful for everyone who would like to start using Cassandra-connector : )

Yes, that's exactly the idea! Good point re the segment size and whether this actually triggers a flush (now you say that, I'm wondering how this is handled in tests of the connector, // CC @bingqinzhou)

@ani-sha, did you actually observe some change events in Kafka after triggering some changes via CQL?

tutorial/debezium-cassandra-init/cassandra.yaml
ani-sha: Point. :sweat_smile:  @gunnarmorling, so I am adding the unchanged config file in this commit. Then add the required changes in a separate commit.

ani-sha:
@gunnarmorling yes, I did changes via cqlsh, and the changes were captured too.

ahmedjami:
Thanks @ani-sha 

From startup-script.sh I can see that you insert new data then you start the connector. In this case, I think that what you observe comes from the snapshot performed by the connector when it starts: `snapshot.mode=ALWAYS`  which allows confirming that the connector was working well : ) 

For the new changes that are written to Cassandra, it's well-explained in the doc that the commit Log will be moved after it reaches the size indicated in cassandra.yaml

ani-sha:
@gunnarmorling Final touch up done! :100:

tutorial/README.md
gunnarmorling: I've pushed a commit for adding some statements (see above).

tutorial/debezium-cassandra-init/cassandra.yaml
gunnarmorling: Ok, thanks, looking great now!

gunnarmorling:
> From startup-script.sh I can see that you insert new data then you start the connector. In this case, I think that what you observe comes from the snapshot performed by the connector when it starts: snapshot.mode=ALWAYS which allows confirming that the connector was working well : )

Yes, that's what I suspect, too. The records I see in the Kafka topics have set `"snapshot":"true"`. I added some instructions to the README in this PR for doing some data changes after connector start-up, and they don't show up as events in the Kafka topic. So as per @ahmedjami is saying, we should reduce the size after which the logs are copied to the CDC folder, so this would happen after a few statements.

ani-sha:
> > From startup-script.sh I can see that you insert new data then you start the connector. In this case, I think that what you observe comes from the snapshot performed by the connector when it starts: snapshot.mode=ALWAYS which allows confirming that the connector was working well : )
> 
> Yes, that's what I suspect, too. The records I see in the Kafka topics have set `"snapshot":"true"`. I added some instructions to the README in this PR for doing some data changes after connector start-up, and they don't show up as events in the Kafka topic. So as per @ahmedjami is saying, we should reduce the size after which the logs are copied to the CDC folder, so this would happen after a few statements.

@gunnarmorling  so `commitlog_segment_size_in_mb ` how much do we reduce to 8 or 16?

jpechane:
@ani-sha I'd say to 1 but it will stil take significant amount of data to fill it.

ani-sha:
@jpechane true that, that's the least we can provide.

gunnarmorling:
Ok, have you taken a look at the repo linked by @ahmedjami? See step 4 in the [README](https://github.com/ahmedjami/debezium-cassandra) for usage of a tool which should produce a sufficient number of changes to trigger a log flush.

ani-sha:
> Ok, have you taken a look at the repo linked by @ahmedjami? See step 4 in the [README](https://github.com/ahmedjami/debezium-cassandra) for usage of a tool which should produce a sufficient number of changes to trigger a log flush.

```
docker-compose -f docker-compose-cassandra.yaml exec cassandra /opt/cassandra/tools/bin/cassandra-stress  write n=100K cl=one -mode native cql3 user=cassandra password=cassandra
```
This is good to execute before `Modify records in the database via Cassandra client` step?

gunnarmorling:
> This is good to execute before "Modify records in the database via Cassandra client" step?

It would really be the "implementation" of this step, i.e. instead of the manually executed statements I added. Can we control which tables this modifies? And is it only inserts? Or other operations, too?

ahmedjami:
If you want the cassandra-stress performs some changes for the same tables in testdb example, I can write a YAML-based profile to do it : )

ahmedjami:
Or replace the actual inventory.cql and use cassandra-stress for all the example.
1. run cassandra-stress, populate some data and start-up the connector to snapshot all the changes.
2- run a second cassandra-stress to make new data changes 

**Note, that we cannot run delete and update cql operations with cassandra-stress ...**

ani-sha:
@ahmedjami Then utimately we would be doing the basic thing as we do now for `update` and `delete`.

ahmedjami:
if you want, you can merge this PR with a first version of Cassandra tutorial that shows users how changes are captured by the connector during the snapshot step then I'll work on a second version that includes cassandra-stress ?

gunnarmorling:
@ahmedjami, yes, that sounds great, let's do that. Thanks a lot, everyone!

PR #140, opened 2021-01-04
Implementing the Saga pattern with Debezium/CDC

gunnarmorling:
# Demo for implementing distributed transactions via the Saga pattern

The idea is similar to the outbox pattern (and eventually, we might have a similar Quarkus extension for Saga): Emit messages for other services via a "message table". Sagas are modeled as a state machine, based on the invidual states of the contained steps (i.e. message flows with other services). Take a look at the README file for steps of running this.

@hpgrahsl, @murphye, here's the PoC I was talking about. Take a look at `OrderPlacementSaga` to see how it'd be used. The example is that of a purchase order service, whose "placeOrder()" transaction should span three services: order, customer (for checking credit balance), and payment (for processing the payment). The latter two are notified via the "saga outbox". If both invocations succeed, the saga will complete and the PO will be moved to status "processed". If one fails, the other invocation will be aborted (by means of sending another message) and once both flows have either failed or aborted, the saga will abort and the order transition to "cancelled".

That's it, almost feels too simple ;) But I think it can be a powerful and most importantly reliable approach for implementing Saga. No dual writes, reliance on availability of remote HTTP services and the like. Feedback welcome :)

What's still missing:

- [x]  Debezium set-up
- [x]  Consuming services; the back channel is simple REST API endpoints atm., whereas later on, Kafka topics would be used
- [x]  Making the message consumer idempotent, i.e. have a message log as in the outbox example, so to identify messages we've processed already and discard them
- [x]  ~~Housekeeping for un-finished Sagas~~ (not going to implement this as part of this demo)


gunnarmorling:
Pushed an update with some basic instructions in the README file describing the basic demo steps. Note that two of the participating services (customer-service, payment-service) don't exist yet. Their responses currently need emulation via manual invocations of the REST API in the order-service. I'm working on adding them next.

gunnarmorling:
@hpgrahsl, @murphye, et al.: pushed several updates and it's in a state now where you could give it a go if you wanted to. All the services are there and it should work end-to-end. See instructions in the README for details. Some more polishing is in order here and there, but the overall flow is there. Thanks for any feedback!

gunnarmorling:
Ok, that's good to review from my pespective now. @rk3rn3r, @Naros, @jpechane, you might find this interesting too. I'm going to write a blog post about this in the coming days, too.

saga/README.md
```diff
@@ -0,0 +1,121 @@
...
+## Running the Example
+
+Build and start up:
+
+```console
+$ mvn clean verify
+```
+
+```console
+$ docker-compose up --build
```
hpgrahsl, line 22: since the compose file needs ENV var setting for DEBEZIUM_VERSION e.g. 1.4 it might be helpful to explicitly point this out here in the README somewhere?

saga/README.md
```diff
@@ -0,0 +1,121 @@
...
+
+Examine the emitted messages for `payment` and `credit-approval` in Apache Kafka:
+
+```console
+$ docker run --tty --rm \
+    --network saga-network \
+    debezium/tooling:1.1 \
+    kafkacat -b kafka:9092 -C -o beginning -q \
+    -f "{\"key\":%k, \"headers\":\"%h\"}\n%s\n" \
+    -t payment.request
```
hpgrahsl, line 47: if I didn't get the saga flow wrong the credit approval should be executed first so you could swap the order of examining the two topics here in the commands?

saga/README.md
```diff
@@ -0,0 +1,121 @@
...
+
+Examine the saga state in the order service's database:
+
+```console
+$ docker run --tty --rm -i \
+        --network saga-network \
+        debezium/tooling:1.1 \
+        bash -c 'pgcli postgresql://orderuser:orderpw@order-db:5432/orderdb'
+
+select * from order.sagastate;
```
hpgrahsl, line 65: this has to be table `purchaseorder.sagastate` right?

saga/README.md
```diff
@@ -0,0 +1,121 @@
...
+
++--------------------------------------+------------------------------------------------------------------------------------------+----------+---------------------------------------------------+-----------------+-----------+
+| id                                   | payload                                                                                  | status   | stepstate                            | type            | version   |
+|--------------------------------------+------------------------------------------------------------------------------------------+----------+---------------------------------------------------+-----------------+-----------|
+| d08b4e43-8522-4539-9aba-de4bb0dc1a8e | {"payment-due":59,"customer-id":456,"order-id":2,"credit-card-no":"xxxx-yyyy-dddd-aaaa"} | COMPLETED  | {"credit-approval":"SUCCEEDED","payment":"SUCCEEDED"} | order-placement | 1         |
```
hpgrahsl, line 70: I see an additional payload field `"type": "REQUEST"` in my demo run?

hpgrahsl: also should the version column entry really be "1" in this case?

saga/README.md
```diff
@@ -0,0 +1,121 @@
...
+
+Alternatively, you also can access pgAdmin on http://localhost:5050 (pgadmin4@pgadmin.org/admin).
+
+Place an order with an invalid credit card number (the payment service rejects any number that ends with "9999"):
+
+```console
+$ http POST http://localhost:8080/orders < requests/place-invalid-order1.json
+```
+
+Observe how the saga's state is `ABORTED`, with the `payment` step in state `FAILED`, and the `credit-approval` first in step in state `ABORTING`, then `ABORTED`.
```
hpgrahsl, line 82: at the sentence end I guess this should just read "first in state aborting, then aborted" right?

saga/README.md
```diff
@@ -0,0 +1,121 @@
+# Saga Pattern
+
+This example demonstrates how to implement the [Saga pattern](https://microservices.io/patterns/data/saga.html) for realizing distributed transactions across multiple microservices, in a safe and reliable way using change data capture.
+
+Similarly to the outbox pattern, this implementation avoids unsafe dual writes to a service's database and Apache Kafka by channeling all outgoing messages through the originating service's database and capturing them from there using CDC and Debezium.
+
+There are three services involved:
+
+* _order-service:_ originator and orchestrator of the Saga
+* _customer-service:_ approves or rejects the customer's credit needed to fulfill an incoming order
```
hpgrahsl, line 10: is this about the customer's credit *card* I guess?

saga/README.md
hpgrahsl: for me it was version 3 for a successful request which is somewhat expected if this reflects to the saga states/transitions somehow?

saga/README.md
```diff
@@ -0,0 +1,121 @@
...
+
+Now place an order which exceeds the credit limit (the customer service rejects any value over 5000):
+
+```console
+$ http POST http://localhost:8080/orders < requests/place-invalid-order2.json
+```
+
+Observe how the saga's state again is `ABORTED`, with the step states set accordingly.
+
+Now stop the payment service and place a valid order again. Observe how the saga remains in state `STARTED`, with the `credit-approval` step in state `SUCCEEDED` and the `payment` step in state `STARTED`.
```
hpgrahsl, line 92: Since it might not be clear to everyone how to stop a single service from the docker compose setup you may want to point this out. also showing how to restart a single service might be helpful ?

hpgrahsl (commented):
so far only reviewed the README plus setup and demo run. code review follows later :)

saga/init-customer.sql
```diff
@@ -0,0 +1,2 @@
+create schema customer;
```
hpgrahsl, line 1: I assume these 3 .sql files to be there just in case someone wants do use them for init purposes? the actual DB schema is created by ORM setting in the `application.properties` right?

saga/order-service/src/main/java/io/debezium/examples/saga/order/facade/KafkaEventConsumer.java
```diff
@@ -0,0 +1,63 @@
...
+import org.eclipse.microprofile.reactive.messaging.Incoming;
+
+import io.debezium.examples.saga.order.event.CreditApprovalEvent;
+import io.debezium.examples.saga.order.event.PaymentEvent;
+import io.debezium.examples.saga.order.saga.OrderPlacementEventHandler;
+
+@ApplicationScoped
+public class KafkaEventConsumer {
+
+    private static final int RETRIES = 5;
```
hpgrahsl, line 25: might benefit from an externalized app config property?

saga/order-service/src/main/java/io/debezium/examples/saga/order/facade/serdes/CreditDeserializer.java
```diff
@@ -0,0 +1,16 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.examples.saga.order.facade.serdes;
+
+import io.debezium.examples.saga.order.event.CreditApprovalEventPayload;
+import io.quarkus.kafka.client.serialization.ObjectMapperDeserializer;
```
hpgrahsl, line 9: cool. didn't know about this serde functionality as part of quarkus

saga/order-service/src/main/java/io/debezium/examples/saga/order/facade/serdes/CreditEventConverter.java
```diff
@@ -0,0 +1,52 @@
...
+                && target.equals(CreditApprovalEvent.class);
+    }
+
+    @SuppressWarnings("rawtypes")
+    @Override
+    public Message<?> convert(Message<?> in, Type target) {
+        IncomingKafkaRecordMetadata metadata = in.getMetadata(IncomingKafkaRecordMetadata.class)
+                .orElseThrow(() -> new IllegalStateException("No Kafka metadata"));
+
+        UUID messageId = UUID.fromString(getHeaderAsString(metadata.getHeaders(), "id"));
```
hpgrahsl, line 38: for kafka record headers I personally think it could be better to use some pre/post fix to very common names like "id" in order to avoid name clashes?

saga/order-service/src/main/java/io/debezium/examples/saga/order/saga/OrderPlacementSaga.java
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    public static OrderPlacementSaga forPurchaseOrder(PurchaseOrder purchaseOrder) {
+        ObjectNode payload = objectMapper.createObjectNode();
+
+        payload.put("order-id", purchaseOrder.id);
+        payload.put("customer-id", purchaseOrder.customerId);
+        payload.put("payment-due", purchaseOrder.paymentDue);
+        payload.put("credit-card-no", purchaseOrder.creditCardNo);
+        payload.put("type", "REQUEST");
```
hpgrahsl, line 45: should there be constants for payload types i.e. "REQUEST" ?

saga/order-service/src/main/java/io/debezium/examples/saga/order/saga/OrderPlacementSaga.java
```diff
@@ -0,0 +1,108 @@
...
+            return new SagaStepMessage(PAYMENT, getPayload());
+        }
+        else {
+            return new SagaStepMessage(CREDIT_APPROVAL, getPayload());
+        }
+    }
+
+    @Override
+    public SagaStepMessage getCompensatingStepMessage(String id) {
+        ObjectNode payload = objectMapper.createObjectNode().put("type", "CANCEL").put("order-id", getOrderId());
```
hpgrahsl, line 62: should there be constants for payload types i.e. "CANCEL" ?

saga/order-service/src/main/resources/META-INF/resources/index.html
```diff
@@ -0,0 +1,256 @@
+<!DOCTYPE html>
```
hpgrahsl, line 1: same as above w.r.t. to auto-generated default index page

saga/order-service/src/main/resources/application.properties
```diff
@@ -0,0 +1,27 @@
+quarkus.datasource.db-kind=postgresql
+quarkus.datasource.jdbc.url=jdbc:postgresql://localhost:5432/orderdb
+quarkus.datasource.username=orderuser
+quarkus.datasource.password=orderpw
+quarkus.hibernate-orm.log.sql=true
+
+quarkus.hibernate-orm.database.generation=drop-and-create
```
hpgrahsl, line 7: this setting (also found in other services) takes care about DB schema gen which is why I commented on need for the .sql init files

saga/register-credit-connector.json
```diff
@@ -0,0 +1,19 @@
...
+    "database.user": "customeruser",
+    "database.password": "customerpw",
+    "database.dbname" : "customerdb",
+    "database.server.name": "dbserver3",
+    "schema.include.list": "customer",
+    "table.include.list" : "customer.outboxevent",
+    "tombstones.on.delete" : "false",
+    "key.converter": "org.apache.kafka.connect.storage.StringConverter",
+    "value.converter": "org.apache.kafka.connect.storage.StringConverter",
+    "value.converter.schemas.enable": "false",
```
hpgrahsl, line 15: since value is plain String this is not needed I guess?

saga/register-payment-connector.json
```diff
@@ -0,0 +1,19 @@
...
+    "database.user": "paymentuser",
+    "database.password": "paymentpw",
+    "database.dbname" : "paymentdb",
+    "database.server.name": "dbserver2",
+    "schema.include.list": "payment",
+    "table.include.list" : "payment.outboxevent",
+    "tombstones.on.delete" : "false",
+    "key.converter": "org.apache.kafka.connect.storage.StringConverter",
+    "value.converter": "org.apache.kafka.connect.storage.StringConverter",
+    "value.converter.schemas.enable": "false",
```
hpgrahsl, line 15: same here, superfluous setting ? because no schema for plain string

hpgrahsl (commented):
I very quickly read through the implementation and think it's a very nice PoC example. just a few minor notes / nitpicks but nothing really worth considering a change for now.

saga/README.md
gunnarmorling: Done.

saga/README.md
gunnarmorling: Yes, indeed.

saga/README.md
gunnarmorling: Good idea.

saga/init-customer.sql
gunnarmorling: No, these files actually are executed (see Docker Compose set-up, they are copied to init.d). The reason being that ORM doesn't _create_ the schema itself, it only _populates_ it with all the required tables, sequences etc.

saga/README.md
```diff
@@ -0,0 +1,121 @@
+# Saga Pattern
+
+This example demonstrates how to implement the [Saga pattern](https://microservices.io/patterns/data/saga.html) for realizing distributed transactions across multiple microservices, in a safe and reliable way using change data capture.
+
+Similarly to the outbox pattern, this implementation avoids unsafe dual writes to a service's database and Apache Kafka by channeling all outgoing messages through the originating service's database and capturing them from there using CDC and Debezium.
```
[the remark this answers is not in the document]
gunnarmorling: Done.

saga/README.md
gunnarmorling: The order of payment and credit approval event flows doesn't really matter; they could happen in either order, or even interlocked.

saga/README.md
```diff
@@ -0,0 +1,121 @@
...
+$ http PUT http://localhost:8083/connectors/credit-outbox-connector/config < register-credit-connector.json
+```
+
+Place an order:
+
+```console
+$ http POST http://localhost:8080/orders < requests/place-order.json
+```
+
+Examine the emitted messages for `payment` and `credit-approval` in Apache Kafka:
```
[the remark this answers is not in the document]
gunnarmorling: Good idea, done!

saga/README.md
gunnarmorling: Good catch! Updated the example with the current payload. Regarding the version, you are right, it should have been 3 as per the state you reviewed. Now it's 2 actually (i.e. one statement less), as the code first did an INSERT + UPDATE when initially persisting the saga state, whereas now it's only the INSERT.

saga/README.md
gunnarmorling: No, it's about the limit you want to grant the customer when ordering. E.g. you might not want to accept an order over 1K for a new customer. I'll replace it with "credit limit", it's a common term.

saga/customer-service/src/main/java/io/debezium/examples/saga/customer/facade/KafkaEventConsumer.java
```diff
@@ -0,0 +1,56 @@
...
+        }).thenRun(() -> message.ack());
+    }
+
+    private String getHeaderAsString(KafkaRecord<?, ?> record, String name) {
+        Header header = record.getHeaders().lastHeader(name);
+        if (header == null) {
+            throw new IllegalArgumentException("Expected record header '" + name + "' not present");
+        }
+
+        return new String(header.value(), Charset.forName("UTF-8"));
```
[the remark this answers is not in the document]
gunnarmorling: Ah, thx!

saga/customer-service/src/main/java/io/debezium/examples/saga/customer/messagelog/MessageLog.java
```diff
@@ -0,0 +1,37 @@
...
+import javax.persistence.EntityManager;
+import javax.persistence.PersistenceContext;
+import javax.transaction.Transactional;
+import javax.transaction.Transactional.TxType;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+@ApplicationScoped
+public class MessageLog {
```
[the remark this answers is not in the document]
gunnarmorling: Theoretically, yes. But then there is no common framework right now where I could put such things (we might add it to our Quarkus extension eventually). Hence I opted for no code sharing, true microservices style ;)

saga/customer-service/src/main/resources/META-INF/resources/index.html
```diff
@@ -0,0 +1,256 @@
+<!DOCTYPE html>
```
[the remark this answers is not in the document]
gunnarmorling: You are right, they are not needed.

saga/order-service/src/main/java/io/debezium/examples/saga/order/saga/OrderPlacementSaga.java
gunnarmorling: Done.

saga/order-service/src/main/java/io/debezium/examples/saga/order/saga/OrderPlacementSaga.java
gunnarmorling: Done.

saga/order-service/src/main/resources/META-INF/resources/index.html
gunnarmorling: Done.

saga/order-service/src/main/resources/application.properties
gunnarmorling: See my other comment, it doesn't create the schema itself unfortunately.

saga/register-credit-connector.json
gunnarmorling: Indeed, that's a left-over from earlier tries.

saga/register-payment-connector.json
gunnarmorling: Removed.

saga/order-service/src/main/java/io/debezium/examples/saga/order/facade/KafkaEventConsumer.java
gunnarmorling: For an actual implementation, yes :)

saga/customer-service/src/main/java/io/debezium/examples/saga/customer/messagelog/ConsumedMessage.java
```diff
@@ -0,0 +1,45 @@
...
+package io.debezium.examples.saga.customer.messagelog;
+
+import java.time.Instant;
+import java.util.UUID;
+
+import javax.persistence.Entity;
+import javax.persistence.Id;
+
+@Entity
+public class ConsumedMessage {
```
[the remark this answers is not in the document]
gunnarmorling: See my comment elsewhere. You are right in general, right now though I don't have a good place for putting these. We might add the message logging functionality to our Quarkus extension eventually.

saga/order-service/src/main/java/io/debezium/examples/saga/order/facade/serdes/CreditEventConverter.java
gunnarmorling: Yeah, good point. I tend to leave it as is for the sake of simplicity, one could re-configure this in the outbox routing SMT.

saga/README.md
gunnarmorling: Yes :)

gunnarmorling:
Pushed a commit for addressing your remarks, @hpgrahsl. Thank you so much for the thorough review. Will merge later today.

saga/README.md
gunnarmorling: To close the loop: I've changed the implementation now to ensure a sequential flow: first credit approval, then payment. Updated the readme accordingly. I'm leaving the code for the optimsistic locking in place, as a good measure of caution.

gunnarmorling:
Thanks a lot for all the comments and feedback, @hpgrahsl et al.! This is in a state now where I think it can be merged. I also added OpenTracing integration so you can examine the Saga flow in Jaeger. The implementation is sequential now, to only do the payment call once after the customer credit approval has succeeded. That should be more realistic in terms of "cost" for compensating/aborting activities after a failure.

The initial timing issue I had was resolved by shorting Debezium's polling timeouts, as pointed out by @jpechane. Go figure :)

hutchig:
I find this really powerful in its conceptual simplicity, some of the code involved I can't really comment on from a standing start with Debezium etc., but I think the way that the saga completes even when the payment service goes down and up is very easy to digest. 

Another concept that might be great to pick out is the way the SagaBase#goBack can peel back a Saga' step's step-by-step without the individual steps' services having to know anything about other actions either siblings or what comes before and after themselves.

On looking at the onMessage->onCreditEvent chain and the similarities between, for example payment KafkaEventConsumer and customer KafkaEventConsumer I was wondering how to abstract the type
differences away and perhaps remove the need for less expert user to write those sorts of intermediate handlers?

PR #141, opened 2021-01-08
[infinispan] Initial import

gunnarmorling:


gunnarmorling:
Hey @karesti, this provides the initial infrastructure for our demo; see the README.md for the details. I'll merge this one right away, so you can take it from there and e.g.

- expand the _cache-update-service_, so that it consumes the events and writes them to Infinispan
- expand the _order-service_ so it has a GET method which retrieves the data from the local cache node

Note there are two topics, purchaseorder and orderline. Ideally, we'd have a single cache, which represents orders as a nested structure, containing their order lines. Do you think that's doable through that cache updater app? Alternatively, we could add a Kafka Streams app in the middle, which does that joining step, so that the cache updater would already receive such nested events. WDYT?

PR #142, opened 2021-01-13
Add Avro examples for Postgres

ckshekhar:
Currently, there are no examples to use Postgres with Avro format. These examples might help beginners like me who want to quickly experiment with Debezium CDC in action.

* Add Postgres example using Apicurio Avro converter
* Add Postgres example using Confluent Avro converter


gunnarmorling:
Hey, thanks for this contribution! It was kinda intentional though to have the Avro set-up only for one of the databases. Multiplying this out for all of the supported connectors doesn't add much value IMO, at the cost of making things more complex and harder to maintain. In that light, I'm going to close this one. Thanks nevertheless for your efforts!

PR #146, opened 2021-01-19
[infinispan] adding xsite replication and protostream

karesti:
@gunnarmorling using CR1 on the client
2 order services
3 clusters 

The protoclasses are in both projects because when the quarkus app starts it loads and puts the generated protobuf in infinispan. I could not make it happen if this is in another jar (like commons jar).


gunnarmorling:
Woohoo, 🎉 !

PR #149, opened 2021-01-20
[infinispan] fix mapping from cache and put common schema

karesti:


distributed-caching/cache-update-service/pom.xml
```diff
@@ -40,6 +40,12 @@
   </build>
 
   <dependencies>
+    <dependency>
+      <groupId>io.debezium.examples.distributedcaching</groupId>
+      <artifactId>distributed-caching-order-service</artifactId>
+      <version>${project.version}</version>
+    </dependency>
```
gunnarmorling, line 47: Wondering whether that's a problem for Quarkus? Perhaps the shared classes should go into a separate module? Or is this what you said doesn't work?

gunnarmorling (approved):
LGTM, just one question inline.

distributed-caching/cache-update-service/pom.xml
karesti: I moved the classes into a data-common module (moved to a package commos as well). and when I tried to compile, I got an arc exception

karesti: I lost the history, but it was an arc deployment exception related to entity manager. so before we enter de debugging mode (might be a bug in quarkus or an improvement to be done), I've gone to the simplest solution of moving int a "commons" package (so can e easily moved after) and depend from the kafka to get the model classes into the classpath instead of having them copy/pasted.
However, the protobuf generator does need to live in both projects. if the class is inside a jar the file won't be generated. maybe also something to be improved either in quarkus or in infinispan protostream

PR #151, opened 2021-01-20
DBZ-2678 Add Vitess Connector tutorial example

keweishang:
https://issues.redhat.com/browse/DBZ-2678

This tutorial example has 2 keyspaces (a.k.a schemas):

- Unsharded keyspace: `customer`, which contains 1 table `customers`.
- Sharded (2 shards) keyspace: `inventory`, which contains 3 tables `products`, `products_on_hand` and `orders`. The shard key is the "product id", that is, the `id` column in the `products` table, the `product_id` column in the `products_on_hand` table, the `product_id` column in the `orders` table.

All tables have the same table definitions as other Debezium connectors' tutorial examples. Also, all tables have already populated with rows.


keweishang:
@gunnarmorling Vitess tutorial example is finally ready :)

gunnarmorling:
Hey @keweishang, excellent, thank you so much!

@ani-sha, could you perhaps take a look at this PR, run through the new instructions in the README and see whether everything works as it should? Thanks!

ani-sha (approved):
LGTM. Works as described in README.

tutorial/debezium-vitess-init/local/create_tables_unsharded_customer.sql
```diff
@@ -0,0 +1,41 @@
...
+    id      INT,
+    next_id BIGINT,
+    cache   BIGINT,
+    PRIMARY KEY (id)
+) comment 'vitess_sequence';
+
+INSERT INTO my_order_seq(id, next_id, cache)
+VALUES (0, 1000, 100);
+
+CREATE TABLE my_customer_seq
```
gunnarmorling, line 23: That's interesting, so no actual sequences in Vitess then?

tutorial/docker-compose-vitess.yaml
```diff
@@ -0,0 +1,37 @@
...
+      - 3888:3888
+  kafka:
+    image: debezium/kafka:${DEBEZIUM_VERSION}
+    ports:
+      - 9092:9092
+    links:
+      - zookeeper
+    environment:
+      - ZOOKEEPER_CONNECT=zookeeper:2181
+  vitess:
```
gunnarmorling, line 17: How about making this an actual cluster with say three nodes, so people can explore the sharding semantics of Vitess?

gunnarmorling (approved):
Very nice, thanks a lot, @keweishang! And thanks for reviewing, @ani-sha.

One comment and one question inline. While not a blocker, the cluster would be nice. WDYT?

tutorial/docker-compose-vitess.yaml
keweishang: This vitess container actually already has multiple shards inside. Inside the container, there are:

- 1 sharded keyspace that has 2 shards, each shard is a MySQL cluster that has 3 processes (1 master, 1 replica and 1 slave)
- 1 unsharded keyspace that is itself a MySQL cluster that has 3 processes (1 master, 1 replica and 1 slave).

So in total, we have 9 MySQL processes (6 for the sharded keyspace, 3 for the unsharded keyspace) inside the container.

tutorial/debezium-vitess-init/local/create_tables_unsharded_customer.sql
keweishang: AFAIK, Vitess uses its own implementation of sequence, which is assisted by such sequence tables, to guarantee global uniqueness of incremental keys. Some more documentation can be found here: https://vitess.io/docs/reference/features/vitess-sequences/

gunnarmorling: Ok, gotcha. Thx!

tutorial/docker-compose-vitess.yaml
gunnarmorling: Oh, I see, that's interesting. I guess there's no way then though to e.g. take down a shard? OTOH, I suppose running nine separate containers might be a tad too much 🤔 .

keweishang: > I suppose running nine separate containers might be a tad too much.

Yeah, running nine processes within a container might be a bit too much. Here's the resource (CPU/MEM) utilization of the containers on my laptop:
```bash
>docker stats
CONTAINER ID   NAME                   CPU %     MEM USAGE / LIMIT     MEM %     NET I/O         BLOCK I/O         PIDS
f33eb73c9013   tutorial_connect_1     2.54%     803.5MiB / 19.56GiB   4.01%     150kB / 118kB   85MB / 106kB      42
7ab005a3059a   tutorial_kafka_1       3.24%     396MiB / 19.56GiB     1.98%     184kB / 221kB   19.6MB / 492kB    73
c00482db3ae8   tutorial_zookeeper_1   0.29%     84.45MiB / 19.56GiB   0.42%     73.1kB / 65kB   49.2MB / 537kB    52
4d54a59ae195   tutorial_vitess_1      45.26%    2.327GiB / 19.56GiB   11.90%    1.33kB / 0B     90.6MB / 1.69GB   727

```

The [official Vitess local docker setup](https://github.com/vitessio/vitess/tree/master/docker/local) also uses multiple processes within one container.

I think the tricky part is that **sharded keyspace**(s) store their sequences in an **unsharded keyspace**. Each shard is itself a MySQL cluster (one master, one replica slave, and one rdonly slave). The vitess docker image used by the Vitess Connector's integration tests also has this setup (9 processes within one container) to have sharded keyspace available for the test.

A few optimized options:

Option 1:
If it's too much, I can reduce the MySQL cluster to have only one master process, therefore, we'd only have in total 3 processes within the container. However, Vitess Connector then has to subscribe to the MySQL master instance, which is not usually the case in a production environment. In production, Vitess Connector usually subscribes to either replica or rdonly slave.

Option 2:
I can further reduce the resource utilization by having only one unsharded space in the example tutorial, which consists of a MySQL cluster of only one master. Therefore, only 1 process running.

Option 3:
Use 9 services in the docker-compose? 3 services for the unsharded keyspace (master, replica, rdonly) and 6 services for the sharded keyspace (master/replica/rdonly for shard No.1, master/replica/rdonly for shard No.2)?

WDYT?

keweishang: >I guess there's no way then though to e.g. take down a shard?

Indeed, if we keep all the processes in the same container (a.k.a docker-compose service), we'll need to `docker exec -it tutorial_vitess_1 bash` into the container to kill all the 3 MySQL processes (master, replica and rdonly) of a given shard.

keweishang: I draw an architecture diagram to help understand the number of processes in the Vitess container. Actually, each square and cylinder represents a process :)
vitess_container

gunnarmorling: Wow, that's really helpful. Thanks a lot, @keweishang!

tutorial/debezium-vitess-init/Dockerfile
```diff
@@ -0,0 +1,25 @@
+# Use a temporary layer for the build stage.
+FROM vitess/base:v8.0.0 AS base
```
gunnarmorling, line 2: Is there away to avoid using this image? It appears to be huge 😱:

vitess/base                                                  v8.0.0                 39997be88418        3 months ago        6.1GB

gunnarmorling (approved):
@keweishang, ok, just gave it a try and things work as they should. I think we can go with what we have (see one comment inline about the humongous base image, but it's not a blocker if there's no way around it for now).

Could you only update the README a bit, so to:

* Embed the nice image you created (e.g. see the README in _../saga_ for an example)
* Explain the sharding set-up (essentially what you've already mentioned here in the PR)
* Describe how to take down a shard by getting a shell within the container and stop one of the processes; this could be informally, just to give people a starting point for what to try out, and they can explore this by themselves

Thanks a lot!

tutorial/debezium-vitess-init/Dockerfile
keweishang: That's a good point! The `vitess/base:v8.0.0 ` docker image serves as a temporary image, I copy two binaries (`/vt/bin/vtctl` and `/vt/bin/mysqlctl`) from this temporary image to the final image. So the final image this Dockerfile creates isn't too big (~1G).

I just asked the Vitess community and see if there's any other way to do this: https://vitess.slack.com/archives/C0PQY0PTK/p1612339740175400

I'll keep you updated here.

keweishang:
@gunnarmorling thanks for the feedback!

I've updated the README and added the 3 sections (the image, sharding setup explanation, and how to take down a shard). Let's wait for 1-2 days for the Vitess community's feedback about the alternatives of copying the 2 binaries from the huge docker image. I'll keep you updated here.

gunnarmorling:
Excellent, thank you so much for your continued efforts here. Looking
forward to merging this soon, it should be a nice way for the community to
try out the new connector.

>

keweishang:
Always a great pleasure to contribute. And I really appreciate your feedback! This series of tutorial examples are very helpful for those who get started with Debezium Connectors. Glad that Vitess Connector has an example now :)

keweishang:
The Vitess maintainers are now aware of this [issue](https://vitess.slack.com/archives/C0PQY0PTK/p1612339740175400). Creating smaller docker images is on their TODOs. I've created a separate DBZ ticket ([DBZ-3053](https://issues.redhat.com/browse/DBZ-3053)) to track it.

I think we can merge this PR for now :)

gunnarmorling:
Ok, excellent, going to merge this one now. And yes, let's hope they come up with a smaller image, I think 6 GB is the biggest one I've ever come across ;) Thanks a lot for this great contribution, @keweishang!

PR #158, opened 2021-02-22
DBZ-3134 Adding example for UI

mdrillin:
Adds example for the debezium UI.  @indraraj - I opened a different PR which incorporates comments indicated on the other PR, moves to ui-demo folder, adds screen shots etc.


gunnarmorling:
Nice one, thanks! Leaving it to @indraraj to take another look and merge.

ui-demo/README.md
```diff
@@ -0,0 +1,97 @@
...
+You can examine change events which occur due to database changes
+
+```shell
+# Open in a new terminal
+# Viewing the change events in the kafka container
+docker exec -it kafka bash
+
+./bin/kafka-console-consumer.sh --bootstrap-server $BOOTSTRAP_SERVERS --topic [TOPIC_NAME] --from-beginning
+
+```
```
indraraj, line 97: I think we should also add the cleanup( `docker-compose down`) section at the end.

indraraj:
Thanks, @mdrillin  for taking it forward. Looks good, you can merge it after adding the cleanup section.

indraraj (approved):

gunnarmorling:
Excellent! Could one of you also look into writing a quick blog post
about the UI? Wouldn't have to be long, explaining the general usage,
some screenshots, etc. @rk3rn3r and @mdrillin, perhaps you can take
the lead on this one? I'd expect it to be similar to what's planned
for docs, but the post could contextualize a bit more and is generally
better in terms of sharing it on social media. Thanks!

ui-demo/README.md
mdrillin: Yeah, good idea.  Added final section

mdrillin:
Added the cleanup section - merging.

PR #161, opened 2021-03-05
DBZ-3141 Oracle SQL to JSON tool

jpechane:


gunnarmorling:
@jpechane, is there still value in adding this, now that we have the full DDL parser?

jpechane:
@gunnarmorling Probably not

gunnarmorling:
Ok, going to close it one then. Thx for your efforts nevertheless! The code will be here in case we want to resurrect it later on.

PR #165, opened 2021-04-06
Regexp restrictied, so patterns like _name are ignored

gerbrand:
Strimzi, Kafka on Kubernetes, also generates records with property-names like _name. The regexp was not restricted enough, so it would return values of that property, instead of the property name.


gunnarmorling (approved):

gunnarmorling:
LGTM, merging. Thx, @gerbrand!

PR #168, opened 2021-04-10
Allow dashboard to work with mysql/mariadb events

gerbrand:
When using debezium-mysql-connector, the context of binlog/streaming events has the value binlog instead of streaming.

Using this change in the dashboard both events are displayed. Arguably, a rewrite rule could in monitoring/debezium-jmx-exporter/config.yml could also work, but using a regular expression seems simpler and less drastic.


gunnarmorling:
Yes, LGTM, merging. Going forward, this will be "streaming" for all connectors, including MySQL, but it's good to support the old name, too. Thx!

PR #169, opened 2021-05-02
Upgrade dependencies of monitoring example

astrohsy:
* Upgrade Grafana, Prometheus, JMX Agent versions to the latest
* Move `jmx_prometheus_javaagent.jar` and conf to `/kafka/etc` from `/kafka`, which is the root of Debezium settings
* Fix docker-compose.yaml and Dockerfiles to manipulate dependency versions only from args in docker-compose.yml


monitoring/debezium-jmx-exporter/Dockerfile
```diff
@@ -1,4 +1,9 @@
-FROM debezium/connect:1.1
+ARG DEBEZIUM_VERSION
```
gunnarmorling, line 1: Thanks, thats much better indeed!

PR #173, opened 2021-06-30
DBZ-1709 [outbox] Upgrade to Debezium 1.6 and Quarkus 2.0

Naros:
https://issues.redhat.com/browse/DBZ-1709

### DIFF
--- a/.github/workflows/outbox-workflow.yml
+++ b/.github/workflows/outbox-workflow.yml
@@ -1,0 +1,26 @@
+name: Build [outbox]
+
+on:
+  push:
+    paths:
+      - 'outbox/**'
+      - '.github/workflows/outbox-workflow.yml'
+  pull_request:
+    paths:
+      - 'outbox/**'
+      - '.github/workflows/outbox-workflow.yml'
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Cache local Maven repository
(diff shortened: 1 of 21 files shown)


gunnarmorling:
@Naros, what's missing to un-draft this one? Btw. related to outbox, the same set of changes could be done to the the Saga example, which uses the outbox extension in all of its three services.

Naros:
I'll take a look at the Saga example and add those to this PR since they're somewhat related in that regard.

gunnarmorling:
Gotcha, makes sense.

Another thing (and feel free to reject it due to scope creep): how about adding a basic GitHub Actions workflow for compiling asorted examples upon PRs and pushes to master? Could even be just the outbox and saga ones to begin with, as you see fit. Compilation itself isn't a great assurance for these examples to work, but it's a first step and we can expand over time. WDYT?

Naros:
You actually read my mind as I had just looked at that Jira on that very topic :P

gunnarmorling:
Hehe, cool. Great minds think alike :)

Naros:
@gunnarmorling I would have expected the workflows to have triggered on this PR since they're included but they were not.  AFAICT I don't think I've missed anything for them, do you see anything I could have missed?

gunnarmorling:
> this will take effect in this repository only once they're merged

Yes, I think that's the case.

gunnarmorling (approved):
LGTM; just merge yourself once you feel it's ready. Thx!

PR #175, opened 2021-07-03
DBZ-1749 Basic CI actions for debezium-examples

Naros:
https://issues.redhat.com/browse/DBZ-1749

There were a few examples that did not have any Maven bits and those do not yet have a workflow component.


gunnarmorling (approved):

gunnarmorling:
Nice one, thx! I think that's good for now; at some point we might consider adding some testing of the Compose set-ups, too. But I think there's bigger fish to fry atm.

Out of curiosity: is there a way to somehow template all those GH Actions, given they are very similar?

PR #181, opened 2021-10-06
Adding BOOTSTRAP_SERVERS  in kafka env variable

indraraj:
This is being used to examine change events [here](https://github.com/debezium/debezium-examples/tree/master/ui-demo#examine-the-change-events)

### DIFF
--- a/ui-demo/docker-compose.yaml
+++ b/ui-demo/docker-compose.yaml
@@ -14,6 +14,7 @@ services:
       - zookeeper
     environment:
       - ZOOKEEPER_CONNECT=zookeeper:2181
+      - BOOTSTRAP_SERVERS=kafka:9092
     networks:
       - ui-network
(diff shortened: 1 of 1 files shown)


rk3rn3r:
I think that would insert the value from the host, not from the Kafka container, but I'm not sure.
@gunnarmorling wdyt?

gunnarmorling:
We show usage of the console consumer via the Kafka container in many places, but usually the bootstrap servers are simply specified in place: _--boostrap-server kafka:9092_. Which README are you referring to, @indraraj? Probably better to adjust it over there.

@rk3rn3r, consumption from the host is notoriously tricky due to the "announced host" business. Hence we tend to show usage of the console consumer in containers (part of the same network), or even via Kafkacat in the Debezium tooling image (my personal preference).

PR #182, opened 2021-10-13
Fix asncdctables.sql - only INSERT if not yet done

tomdw:
Make sure INSERT statements are only executed when it is not yet done. Otherwise more than 1 record ends up in e.g. the ASNCDC.IBMSNAP_CAPPARMS table which makes starting the capture process fail on next reboot of the container. Result: process does not start anymore and capture does not work anymore.


gunnarmorling:
@jcechace, perhaps you can take a look into this one?

PR #185, opened 2021-11-17
Update outbox README.md

danialgood:
Updated config syntax for dev mode Datasource since the previous one is deprecated.


gunnarmorling:
Good catch, thx!

PR #189, opened 2021-11-22
[Draft] DBZ-4302 Added an example for MongoDB Outbox Event Router

sgc109:
https://issues.redhat.com/browse/DBZ-4302

---

follow-up for debezium/debezium#2877

### DIFF
--- a/mongodb-outbox/README.md
+++ b/mongodb-outbox/README.md
@@ -1,0 +1,33 @@
+# Outbox Pattern with MongoDB
+
+```shell
+# Start the topology as defined in https://debezium.io/docs/tutorial/
+export DEBEZIUM_VERSION=1.8
+docker-compose -f docker-compose.yaml up --build
+
+# Initialize MongoDB replica set and insert some test data
+docker-compose -f docker-compose.yaml exec mongodb bash -c '/usr/local/bin/init-inventory.sh'
+
+# Start MongoDB connector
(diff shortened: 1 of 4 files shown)


gunnarmorling (commented):
Very nice, this looks great. We only should update to the tagged version once 1.8.0.Beta1 is out (planned for Thursday this week). And did you actually test the tracing integration? As we don't start a span in the source application, the tracing span should start at the point when the transaction got committed.

gunnarmorling:
@sgc109, which MongoDB version are you using? I'm getting this error in the mongo logs:

```
mongodb_1    | {"t":{"$date":"2021-11-22T09:26:00.906+00:00"},"s":"I",  "c":"-",        "id":4939300, "ctx":"monitoring-keys-for-HMAC","msg":"Failed to refresh key cache","attr":{"error":"NotYetInitialized: Cannot use non-local read concern until replica set is finished initializing.","nextWakeupMillis":6800}}
```

Did you encounter that too?

sgc109:
@gunnarmorling, Yes, I encounter that too, it seems to occur literally because replica set is not initialized as it says, so it would have to stop to occur right after you run the next command in README, which executes shell script including replica set initialization process in docker container.

gunnarmorling:
Thanks a lot, @sgc109. Rebased and applied. Pushed one more commit with a few minor adjustments. That issue above was a stupid user mistake on my end; all looking good now. Thanks again for this great contribution!

PR #192, opened 2021-12-25
update docker-compose-postgres.yaml for create connector

sinanbaran:
Postgres wal_level=logical needs to be set to config .https://www.postgresql.org/docs/9.6/runtime-config-wal.html


gunnarmorling:
This is already done via the postgres.conf of the example image. Going to close this one therefore, thanks for your efforts nevertheless!

PR #198, opened 2022-02-02
DBZ-4666 - Add kafkastore.bootstrap.servers configs to schema-registry containers

eric-weaver:
Adds kafkastore.bootstrap.servers to the schema-registry container compose configs. This config option is required in [confluentinc/cp-schema-registry:7.0.0](https://docs.confluent.io/platform/7.0.0/schema-registry/installation/config.html#kafkastore-connection-url) which is being pulled in when the avro examples are run without a local cached version of an older image.  For the purpose of these standalone examples I believe it should be ok to keep the zookeeper connection config around so older versions will still work, but happy to clean it up if it is desired.

### DIFF
--- a/cloudevents/docker-compose.yaml
+++ b/cloudevents/docker-compose.yaml
@@ -52,6 +52,7 @@ services:
       - 8181:8181
       - 8081:8081
     environment:
+      - SCHEMA_REGISTRY_KAFKASTORE_BOOTSTRAP_SERVERS=kafka:9092
       - SCHEMA_REGISTRY_KAFKASTORE_CONNECTION_URL=zookeeper:2181
       - SCHEMA_REGISTRY_HOST_NAME=schema-registry
       - SCHEMA_REGISTRY_LISTENERS=http://schema-registry:8081
--- a/end-to-end-demo/docker-compose.yaml
+++ b/end-to-end-demo/docker-compose.yaml
@@ -36,6 +36,7 @@ services:
      - 8181:8181
      - 8081:8081
     environment:
+     - SCHEMA_REGISTRY_KAFKASTORE_BOOTSTRAP_SERVERS=kafka:9092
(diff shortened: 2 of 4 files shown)


gunnarmorling:
Hey @eric-weaver, thanks a lot (and welcome back, has been a while :)). So this change is good, and we could merge it as-is. Buuut, I'm wondering whether you'd be up for slightly expanding the scope (I know, I'm greedy, ahaha). There's a few related things that would be great to do:

* Pin the version of the registry image
* Replace the equally obsolete option from kafka-topics.sh invocations (see [DBZ-4660](https://issues.redhat.com/browse/DBZ-4660)
* Update the four affected examples to Debezium 1.8 (i.e. version in READMEs essentially) and verify that the given instructions still work; some of the examples are severely outdated also otherwise (e.g. the end-to-end one), but if we just could move them to the latest Debezium (and thus Kafka 3.0), that'd be great

Again, no pressure at all, just gauging whether you'd be up for some or all of these. No problem at all if not :)

gunnarmorling:
Applied. Thanks a lot, @eric-weaver! I've captured the proposed follow-up changes in DBZ-4660, up for anyone to grab.

eric-weaver:
Hey @gunnarmorling, nice to see you remember! It has been a while, but we've got some heavy focus on some expanding our CDC pipeline for outbox so I'm sure you'll see more from us :). Thank you for accepting this change. I would like to contribute more back to this project so I will try and see if I can get to these changes later next week, but will leave the ticket for anyone to pickup in the meantime if they desire!

gunnarmorling:
> so I'm sure you'll see more from us :)

Nice, looking forward to it. Also, should you still be interested in that idea of creating a guest post about your experiences with CDC and Debezium, let me know. Outbox would definitely be interesting to folks out there.

> so I will try and see if I can get to these changes later next week, but will leave the ticket for anyone to pickup in the meantime if they desire!

👍

PR #204, opened 2022-02-10
DBZ-4440 Use quay.io container registry and update image versions

vjuranek:
Update versions of the docker images and use quay.io container registry instead of docker hub. Where needed, fix the example to work with recent images, see [DBZ-4440](https://issues.redhat.com/browse/DBZ-4440)

There are still some issues, which are tracked by separate Jiras:
* Openshift example wasn't updated as this example is outdated and needs to be re-worked, [DBZ-1994](https://issues.redhat.com/browse/DBZ-1994)
* ElasticSearch in jpa-aggregations examples doesn't work properly, [DBZ-4788](https://issues.redhat.com/browse/DBZ-4788)


gunnarmorling:
Thanks a lot, @vjuranek! Two thoughts:

* We currently lack means of regression testing for the examples, i.e. any update could potentially break things; I'm not so concerned about the 1.7 -> 1.8 update, those should be fine. But could you run through the instructions for those with the large bumps, like 0.9 -> 1.8? Some more changes may be needed there, in particular as we did some clean-up in 1.0
* Could you attribute the commit either to that Jira issue with the schema registry fix (thus widening its scope a bit), or a new issue for that version update?

vjuranek:
>     * We currently lack means of regression testing for the examples, i.e. any update could potentially break things; I'm not so concerned about the 1.7 -> 1.8 update, those should be fine. But could you run through the instructions for those with the large bumps, like 0.9 -> 1.8? Some more changes may be needed there, in particular as we did some clean-up in 1.0

Yes, I took it into account. Actually I skipped couple of place where I expect some issues and will address it separately. In this PR I bump some outdated versions in
* apache-pulsar - seem broken regardless of DB image update, looking on it, will send another PR soon
* nat - DB image update works, unfortunately I tested it before bumping Debezium version (which I consider to be safe, so I missed the compilation issue:-()
* openshift - [register.json](https://github.com/debezium/debezium-examples/blob/main/openshift/register.json) looks correct, so I consider this as a safe change, but anyway, I'll try to check today and will update here
 
>     * Could you attribute the commit either to that Jira issue with the schema registry fix (thus widening its scope a bit), or a new issue for that version update?

ack

> 
> _Update:_ at least the NATS example will need some more work, the CI build (we have that for some chosen ones, but it's only about compilation, not actually running the examples) breaks due to the recent migration from log4j to logback.

sorry for that, I missed it, see above, will fix it soon

vjuranek:
apache-pulsar, nats and openshift removed from which PR. Now it contains only bump 1.7.1 -> 1.8.1

vjuranek:
there will be more changes, converting to draft

vjuranek:
Updated PR summary.

gunnarmorling:
Rebased and applied. Great work, @vjuranek, thanks a lot! This update was long overdue, great to see it in place.

PR #205, opened 2022-02-11
DBZ-4738 Make nats example working with recent Debezium

vjuranek:
DBZ-4738 Remove nats example
    
Nats is intergared into Debezeim as part of Debezium server and we
already has example for Debezium server. This example is not needed
any more. Remove it.

https://issues.redhat.com/browse/DBZ-4738


gunnarmorling:
Hey @vjuranek, so there's something I'm just realizing now (sorry for not thinking of this earlier): Since this example was created, [support for NATS](https://debezium.io/documentation/reference/stable/operations/debezium-server.html#_nats_streaming) was added to Debezium Server. So in that light, the example in its current form (based on the embedded engine) doesn't add that much value any longer and I think in fact we may just delete that example. 

Alternatively, we could remodel the NATS example to be based on Debezium Server, but we already have other Server examples, e.g. for [Pub/Sub](https://github.com/debezium/debezium-examples/tree/main/debezium-server-mongo-pubsub), so value of that isn't super high either probably. My vote is on removing the example.

WDYT?

vjuranek:
I vote also for removing - less code means less bugs and less outdated code in the future

gunnarmorling:
Ok, sweet. So could you rework that PR then for dropping that example? Thanks!

vjuranek:
> Ok, sweet. So could you rework that PR then for dropping that example? Thanks!

done

gunnarmorling:
Applied, thx!

PR #213, opened 2022-03-08
DBZ-1994 Update OpenShift example to OpenShift 4

vjuranek:
https://issues.redhat.com/browse/DBZ-1994


vjuranek:
@gunnarmorling do you want me to add k8s secret into this example as well or keep it as simple as possible (and eventually link the documentation how to do that)? Personally I would keep it simple, as adding secret is possible source of issues when user start playing with it (it took me quite some time to figure out that encoding login and password with `echo 'passoword'|base64` returns wrong encoded value because `echo` adds new line to the string, and there are probably more possible pitfalls like this).

vjuranek:
Updated example also with secret configuration.

vjuranek:
Updated example also with secret configuration.

gunnarmorling:
Excellent, thanks a lot!

PR #215, opened 2022-03-29
Correct parameter KAFKA_CONNECT_URIS

a10pepo:
As per documentation:

https://debezium.io/documentation/reference/stable/operations/debezium-ui.html

The correct naming for the parameter is KAFKA_CONNECT_URIS


gunnarmorling:
Thx!

PR #216, opened 2022-04-08
DBZ-1013 Automatically download Oracle JDBC driver from Maven Central

Naros:
https://issues.redhat.com/browse/DBZ-1013

This is a repurpose effort on DBZ-1013 based on our discussions on Zulip to automatically download & install the Instant Client rather than relying on the user to do this step.  This should simplify the steps needed by the user when getting started with Oracle via the tutorial and aligns this closer to other database vendors.


gunnarmorling:
Yeah, that's a nice improvement. If we were to solely show LogMiner in the tutorial example, could we then do away with Instantclient completely and just get the driver via curl/wget from Maven Central?

Naros:
Yes, is that the route you'd prefer to take, just removing Xstream entirely?

gunnarmorling:
Yeah, I think that'd be preferable.

tutorial/debezium-with-oracle-jdbc/Dockerfile
```diff
@@ -1,14 +1,27 @@
 ARG DEBEZIUM_VERSION
 FROM quay.io/debezium/connect:$DEBEZIUM_VERSION
 ENV KAFKA_CONNECT_JDBC_DIR=$KAFKA_CONNECT_PLUGINS_DIR/kafka-connect-jdbc
-ENV INSTANT_CLIENT_DIR=/instant_client/
+
+# These define where we will automatically fetch the Oracle Instant Client from
+# This uses version 21.1.0.0, and all these should remain in-sync version wise with one another.
+ENV INSTANT_CLIENT_RPM_BASE_URL=https://yum.oracle.com/repo/OracleLinux/OL8/oracle/instantclient21/x86_64/getPackage/
```
jpechane, line 7: One question - would it be possible to add this as a repository using microdnf so it would be installed using the standard way?

gunnarmorling: So you want to keep XStream in that example after all?

tutorial/debezium-with-oracle-jdbc/Dockerfile
```diff
@@ -1,14 +1,17 @@
...
+    ORACLE_JDBC_REPO="com/oracle/database/jdbc" \
+    ORACLE_JDBC_GROUP="ojdbc8" \
+    ORACLE_JDBC_VERSION="21.1.0.0" \
+    ORACLE_JDBC_MD5=ebb69c8de3a83c08846b3fd47836ff15
+
+RUN docker-maven-download central "$ORACLE_JDBC_REPO" "$ORACLE_JDBC_GROUP" "$ORACLE_JDBC_VERSION" "$ORACLE_JDBC_MD5"
 
-USER kafka
-# Deploy Oracle client and drivers
+USER root
```
jpechane, line 14: Is the `libaio` now needed? If not then let's remove it.

jpechane:
@Naros Applied, thanks

PR #217, opened 2022-04-19
[Distributed Caching] Optimistic locking support;

gunnarmorling:
When writes are issued based on a stale cache read, they will be rejected.


vjuranek (approved):

vjuranek:
Merged, thanks @gunnarmorling !

PR #219, opened 2022-05-01
Fix dockerfile using microdnf instead of yum

adolsalamanca:
### Description
* Currently, dockerfile for kstreasm example is not working due to Debezium docker image is using fedora minimal image, which does not include yum.
* The purpose of this tiny PR is to fix that replacing it for microdnf.


jpechane:
@adolsalamanca Applied, thanks!

PR #223, opened 2022-05-20
To make Kafka accessible from IDE on port 9093

kmozaid:


jpechane:
@kmozaid Hi, thanks for the PR

- Is it ok to advertise endpoint that is not accessible for some of the clients?
- WOuld it be possible to introduce the same change into all tutorial composes?

kmozaid:
> @kmozaid Hi, thanks for the PR
> 
> * Is it ok to advertise endpoint that is not accessible for some of the clients?
> * WOuld it be possible to introduce the same change into all tutorial composes?

HI @jpechane 

- The endpoint `EXTERNAL://localhost:9093` is advertised so that clientson docker host can access Kafka running in docker. When we specify endpoint for clients running on docker host, we also need to advertise endpoint for internal service like kafka connect (The endpoint `CLIENT://kafka:9092` is for Kafka Connect Docker).
- Sure, I can update other docker-compose files.

rk3rn3r:
Quick question: Is it valuable to map to a static port? Won't this conflict with multiple containers running concurrently? Would it be possible to use random ports and map the config in the entrypoint script?
Else, we should at least mind in the docs that this can conflict and running the same container multiple times isn't supported.
wdyt @jpechane ?

tutorial/docker-compose-postgres.yaml
```diff
@@ -10,10 +10,15 @@ services:
     image: quay.io/debezium/kafka:${DEBEZIUM_VERSION}
     ports:
      - 9092:9092
```
OneCricketeer, line 12: Since 9092 is only advertised in the Docker network, and not reachable from the host, this should be removed

OneCricketeer:
> Won't this conflict with multiple containers running concurrently

Yes.

> possible to use random ports and map the config in the entrypoint script

`advertised.listeners` needs to have a static value, albeit one could dynamically generated, but then how would compose or the Dockerfile expose that random value without defining it statically in those configs?

tutorial/docker-compose-postgres.yaml
kmozaid: done. thanks

kmozaid:
@OneCricketeer and @jpechane  Sorry for the very late update on this PR, please review once again, I have made the required changes.

PR #281, opened 2022-10-11
[Kinesis] Switch to Debezium 2.0

Naros:


Naros:
@jpechane this is ready.

PR #294, opened 2022-10-13
[Camel-component] Switch to Debezium 2.0

vjuranek:
This is only partial fix, as Camel doesn't support Debezium 2.0.0 yet, so the demo fails with wrong configuration of camel-debezium-postgres.

See https://issues.redhat.com/browse/DBZ-5725 and
https://issues.apache.org/jira/browse/CAMEL-18610

### DIFF
--- a/.github/workflows/camel-component.yml
+++ b/.github/workflows/camel-component.yml
@@ -15,6 +15,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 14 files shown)


vjuranek:
This should finally work. CI fails as the Camel 4.0.0 (Camel version with Debezium 2.0 suport) requires Java 17 while we compile on the CI with Java 11.

mfvitale (approved):

PR #298, opened 2022-10-14
[Debezium Server NameMapper PubSub] Switch to Debezium 2.0

jcechace:
This one was failing with [missing method](https://github.com/a0x8o/kafka/blob/master/clients/src/main/java/org/apache/kafka/common/utils/Utils.java#L1442). This method is not present in kafka-clients:3.1 (which is defined in quarkus' BOM). The exclusion and manual alignment seems to do the trick, although I'm not sure if this should be the solution

### DIFF
--- a/debezium-server-name-mapper/README.md
+++ b/debezium-server-name-mapper/README.md
@@ -20,7 +20,7 @@ Both Apache Pulsar and the source database are deployed via Docker Compose file.
 From terminal start the source database and the sink system:
 
 ```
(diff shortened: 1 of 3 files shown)


jpechane:
@jcechace We have 2.0.0.Final availble. Could you please retry with it?

jcechace:
@jpechane  done. I reverted the exclusion since Final doesn't seem to need it.

PR #300, opened 2022-10-14
[MongoDB Outbox] Switch to Debezium 2.0

jcechace:


jpechane:
@jcechace 2.0.0.Final is available

PR #307, opened 2022-11-03
Fix markdown typo

syl-oh:
Hello I was reading the doc and saw that there was a missing `]` there.


jpechane:
@syl-oh Applied, thanks

PR #314, opened 2023-03-14
Update Prometheus JMX Exporter config rules for mult-partition connector mode

dude0001:


dude0001:
Should I target 2.x branch instead of main?

jpechane:
@dude0001 No, the `main` is correct. The `2.x` branch was used for 1.x -> 2.x migration.

jpechane:
@dude0001 Thanks for the PR. I did a force push to update commit message.
Will this update work for non-partitioned connectors - like MySQL too? If not could you please modify it so it works for both types?

Naros:
I recently updated the monitoring example to support multi-partition and non-partitioned connectors, so I believe this PR is not obsolete.  @dude0001 could you check and verify if that's accurate and close if so?

dude0001:
Your change looks good to me @Naros, thank you for updating it.

PR #315, opened 2023-03-20
DBZ-6216 Use quay.io image

vjuranek:
https://issues.redhat.com/browse/DBZ-6216


jpechane:
@vjuranek Applied, thanks

PR #320, opened 2023-04-14
DBZ-6338  Added an example offset editor application to allow editing…

nathan-smit-1:
https://issues.redhat.com/browse/DBZ-6338

### DIFF
--- a/offset-editor/src/main/java/io/debezium/examples/offset/editor/OffsetEditorApp.java
+++ b/offset-editor/src/main/java/io/debezium/examples/offset/editor/OffsetEditorApp.java
@@ -1,0 +1,218 @@
+package io.debezium.examples.offset.editor;
+
+import java.io.*;
+import java.nio.ByteBuffer;
+import java.nio.charset.StandardCharsets;
+import java.util.HashMap;
+import java.util.Map;
+import javafx.application.Application;
+import javafx.collections.FXCollections;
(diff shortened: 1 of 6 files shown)


offset-editor/src/main/java/io/debezium/examples/offset/editor/OffsetEditorApp.java
```diff
@@ -0,0 +1,218 @@
...
+        primaryStage.setTitle("Offset Editor");
+        primaryStage.setScene(scene);
+        primaryStage.show();
+    }
+
+    private FileChooser createFileChooser() {
+        FileChooser fileChooser = new FileChooser();
+        fileChooser.setTitle("Select offset.dat File");
+        fileChooser.getExtensionFilters().add(
+                new FileChooser.ExtensionFilter("Offset File", "*.dat"));
```
jpechane, line 59: I don't think we should limit to only `dat` extension as it can be arbitrary.

jpechane (commented):
@nathan-smit-1 Thanks for the PR, I think this is great start.

I left few comments but I've one more general proposal. In GUI when file is opened you can eddit the JSON for a selected key. Would it be possible to list the keys and then whe a key is selected then edit that JSON as a table?

jpechane:
@nathan-smit-1 @ani-sha LGTM! Thanks

PR #324, opened 2023-07-08
Update the way environment variables are set in Docker compose file

ldy9037:
Hi! 
First, I'm not familer to English, so I apologize for the awkwardness.

While following the tutorial, I noticed that export DEBEZIUM_VERSION doesn't apply.
And I found out that I need to set up a separate environment variable file (.env) to solve this problem.
What do you think about updating the README.md document to clear up confusion when running the tutorial?

Thanks!

### DIFF
--- a/tutorial/README.md
+++ b/tutorial/README.md
@@ -26,7 +26,8 @@ This demo automatically deploys the topology of services as defined in the [Debe
 
 ```shell
 # Start the topology as defined in https://debezium.io/documentation/reference/stable/tutorial.html
-export DEBEZIUM_VERSION=2.1
+
+echo DEBEZIUM_VERSION=2.1 > .env
 docker-compose -f docker-compose-mysql.yaml up
 
 # Start MySQL connector
@@ -196,7 +197,7 @@ docker run --rm --tty \
 
 ```shell
 # Start the topology as defined in https://debezium.io/documentation/reference/stable/tutorial.html
-export DEBEZIUM_VERSION=2.1
+echo DEBEZIUM_VERSION=2.1 > .env
 docker-compose -f docker-compose-postgres.yaml up
 
 # Start Postgres connector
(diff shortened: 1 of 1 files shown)


jpechane:
@ldy9037 Hi, thanks for the PR. Before we apply it, could you please elaborate more on `export DEBEZIUM_VERSION` does not apply? In what sitation do you need it and it is not available?

ldy9037:
Thanks for the quick reply!

I use "/debezium-examples/tutorial/docker-compose-oracle.yaml" to test Debezium for Oracle.
I've tried like this...
```sh
# OS : ubuntu 20.04
# Docker : 24.0.2
# docker-compose : 1.25.0

$ export DEBEZIUM_VERSION=2.1
$ echo $DEBEZIUM_VERSION
2.1
$ sudo docker-compose -f docker-compose-oracle.yaml up --build
WARNING: The DEBEZIUM_VERSION variable is not set. Defaulting to a blank string.
ERROR: invalid reference format
```
I wanted the containers defined in the docker-compose file to be created normally.
So I tried again using ".env"
```sh
$ echo DEBEZIUM_VERSION=2.1 > .env
$ sudo docker-compose -f docker-compose-oracle.yaml up --build
Pulling zookeeper (quay.io/debezium/zookeeper:2.1)...
2.1: Pulling from debezium/zookeeper
...
```

Thanks!

ldy9037:
@mfvitale All of the methods you suggested worked for me! (I'm using bash shell)
I learned that environment variables are not kept when using the sudo command and I also learned that the -E flag should be used when environment variables are to be keep.
- https://www.sudo.ws/docs/man/1.8.22/sudo.man/#E

And to prevent unnecessary sudo use, I decided to follow the manual you suggested.
- [Manage Docker as a non-root user](https://docs.docker.com/engine/install/linux-postinstall/#:~:text=If%20you%20don't%20want,members%20of%20the%20docker%20group)
- [Manage Docker as a non-root user](https://docs.docker.com/engine/security/rootless/ ) (For security)

I believe the issue is related to using Docker and not the Debezium tutorial. Therefore, this Pull Request is considered unnecessary. What do you think we should do? If you agree with me, please close this pull request.

Thanks!

mfvitale:
Closing because is not Debezium related.

PR #326, opened 2023-08-01
[http-signaling-notification] Use Debezium Final version instead of SNAPSHOT

mfvitale:


ani-sha (approved):

jpechane:
@mfvitale Applied, thanks

PR #328, opened 2023-08-04
Update Examples for Debezium 2.3

Naros:


jpechane:
I wonder if we can come up with a solution to define the version in a single place. It will be probably quite hard for Dockerfiles (we might need to introduce a script that will do it) but java based examples might benefit from a common parent. WDYT?

Naros:
I think it's partially doable:

1. Maven could import or be based on some `debezium-examples-parent` pom.xml
2. The docker compose files could use `extends`, but would require docker-compose v3
3. Docker files could potentially just point to `<image>:latest` tbh.

All that would then remain would be the `README.md` and other files that directly reference the versions.  Perhaps we might be able to get away with using `latest` in these contexts too?

Naros:
Superseded by #404

PR #330, opened 2023-09-12
DBZ-6893 Migrated all MongoDB examples to mongodb.connection.string

jcechace:
https://issues.redhat.com/browse/DBZ-6893


jcechace:
Ready for merge. I wasn't able to make the mongodb outbox works -- however that was also the case for the former version so the issue (unless it's between the chair and the keyboard) is unrelated. I'm looking into what might be the cause there.

PR #331, opened 2023-09-13
[tensorflow] Move example into dedicated ML folder

vjuranek:
As more ML examples will be added, let's have all the example in the same sub-folder.


vjuranek:
Closing, integrated into PR #332.

PR #334, opened 2023-09-27
DBZ-4471 Introducing Debezium Server Example with MySQL and GCP PubSub

ilyasahsan123:
https://issues.redhat.com/browse/DBZ-4471


debezium-server/debezium-server-sink-pubsub/config-mysql/application.properties
```diff
@@ -0,0 +1,12 @@
+debezium.sink.type=pubsub
```
jpechane, line 1: Did you try to run it? Does it really work without internal schema history configured?

jpechane:
@ilyasahsan123 Thanks for the PR, could you please remove the brackets around `DBZ-4471` in commit messages?

debezium-server/debezium-server-sink-pubsub/config-mysql/application.properties
ilyasahsan123: Edited. I also tested the connector with the result event was successfully sent to Pubsub.

ilyasahsan123:
> @ilyasahsan123 Thanks for the PR, could you please remove the brackets around `DBZ-4471` in commit messages?

Sure. edited.

debezium-server/debezium-server-sink-pubsub/README.md
```diff
@@ -0,0 +1,125 @@
...
+```
+
+Test the setup by making changes to the customers table. The logs will appear in Google Cloud Pub/Sub shortly. You can access Postgres with this command:
+
+```shell
+docker compose exec postgres env PGOPTIONS="--search_path=inventory" bash -c 'psql -U $POSTGRES_USER postgres'
+```
+
+
+### MongoDB Debezium Connector
```
jpechane, line 85: Please take a look at original MongoDB documentation. The setup sequence is different from MySQL or Postgres. Also this text speaks about database and table which is not correct in context of MongoDB.

jpechane (changes requested):
@ilyasahsan123 Thanks a lot. IMHO we are almost done. There is just one remaining issue in MongoDB documentation.

debezium-server/debezium-server-sink-pubsub/README.md
ilyasahsan123: Thanks for your feedback,@jpechane. Could you please re-review this?

jpechane:
@ilyasahsan123 Applied, thanks a lot

PR #341, opened 2024-01-21
DBZ-7370 PubSub example for DS deployed via operator

jcechace:
https://issues.redhat.com/browse/DBZ-7370

Will be also used as a demo for Quarkus Insights


jpechane:
@jcechace Applied, thanks!

PR #344, opened 2024-02-01
parameterize pg version

shivam01anand:


jpechane:
I don't think this is correct. The exmaple is based on assumption that the POstgres database is empty and data are replicated from MySQL. Image from examples has already data present. If a parameter is going to be introduced it has to be a new one with postgres version.

PR #345, opened 2024-02-05
[Debezium Server Name Mapper] Switch to Debezium 2.5, Quarkus to 3.0.0.Final

ani-sha:


debezium-server-name-mapper/pom.xml
```diff
@@ -7,12 +7,12 @@
...
-        <maven.compiler.target>1.8</maven.compiler.target>
+        <maven.compiler.source>11</maven.compiler.source>
+        <maven.compiler.target>11</maven.compiler.target>
 
-        <version.debezium>2.1.3.Final</version.debezium>
-        <version.debezium.tag>2.1</version.debezium.tag>
-        <version.quarkus>2.16.5.Final</version.quarkus>
+        <version.debezium>2.5.0.Final</version.debezium>
+        <version.debezium.tag>2.5</version.debezium.tag>
+        <version.quarkus>3.0.0.Final</version.quarkus>
```
jpechane, line 15: Debezium 2.5 uses Quarkus 3.2.3.Final. Maybe it i better to align it here?

jpechane:
@Naros With this upgrade it seems the example no longer works. It seems that stream mapper bean is no longer resolved/found. Do you know if there was an change in the lookup mechanism?

Naros:
Ya the issue here was that the pom was referencing the old Java EE cd-api rather than the Jakarta EE cdi-api that should be used now-a-days with the major shift from Java EE to Jakarta EE.  I've added a commit that solves the issue and verified that the output, as seen here for the first entry on the query:

```
┌─ccranfor@fedora ~/github/Naros/debezium-examples-4/debezium-server-name-mapper (dbz-server-name-mapper-update)
└─$ docker compose exec pulsar bin/pulsar-admin broker-stats topics -i
{
    "public/default": {
        "0x00000000_0x40000000": {
            "persistent": {
                "persistent://public/default/superprefix.tutorial.inventory.products_on_hand": {
```

Naros:
Thanks @ani-sha, applied!

PR #346, opened 2024-02-06
fix(): typo in folder name

pneff93:


jpechane:
@pneff93 Applied, thanks

PR #354, opened 2024-04-27
DBZ-7826 Abnormal Behavior in Debezium Monitoring Example

chungeun-choi:
https://issues.redhat.com/projects/DBZ/issues/DBZ-7826?filter=addedrecently


jpechane:
@chungeun-choi Applied, thanks

PR #360, opened 2024-09-26
Update unwrap-smt example to use Debezium JDBC sink connector

vjuranek:


Naros (commented):
Thanks @vjuranek, applied.

PR #370, opened 2024-12-06
Cleanup: Remove duplicate topics key

overwatcheddude:


jpechane:
@overwatcheddude Applied, thanks for spotting.

PR #372, opened 2024-12-11
DBZ-1937: Add example for SSL-enabled Kafka

Sgitario:
Added a new example to explain how to use Debezium connector with a SSL-enabled Kafka instance.

Relates https://issues.redhat.com/browse/DBZ-1937


kafka-ssl/README.md
```diff
@@ -0,0 +1,133 @@
...
+      - kafka
+      - postgres
+```
+
+Note that we're using the network mode as "host", because the certificates only validate the hostname "localhost", so the kafka and connector instances must be on the same network. We can also disable the hostname validation as an alternative.
+
+## Steps
+1. Start the topology (postgresql, kafka, and debezium connector) as defined in https://debezium.io/documentation/reference/stable/tutorial.html:
+```shell
+export DEBEZIUM_VERSION=2.1
```
jpechane, line 67:
```suggestion
export DEBEZIUM_VERSION=3.0
```

kafka-ssl/README.md
```diff
@@ -0,0 +1,133 @@
...
+      - STATUS_STORAGE_TOPIC=my_connect_statuses
+      - CONNECT_SECURITY_PROTOCOL=SSL
+      - CONNECT_SSL_KEYSTORE_LOCATION=/opt/config/ssl/kafka-ssl-keystore.p12
+      - CONNECT_SSL_KEYSTORE_PASSWORD=top-secret
+      - CONNECT_SSL_KEYSTORE_TYPE=PKCS12
+      - CONNECT_KEY_PASSWORD=top-secret
+      - CONNECT_SSL_TRUSTSTORE_LOCATION=/opt/config/ssl/kafka-ssl-truststore.p12
+      - CONNECT_SSL_TRUSTSTORE_PASSWORD=top-secret
+      - CONNECT_SSL_TRUSTSTORE_TYPE=PKCS12
+    network_mode: host
```
jpechane, line 54: Why is this necessary?

kafka-ssl/register-postgres.json
```diff
@@ -0,0 +1,26 @@
+{
+    "name": "inventory-connector",
+    "config": {
+        "connector.class": "io.debezium.connector.postgresql.PostgresConnector",
```
jpechane, line 4: Please use MySQL connector. PostgreSQL connector does not need internal schema history so the lines below would be irelevant.

kafka-ssl/README.md
```diff
@@ -0,0 +1,133 @@
...
+The certificates will be mounted at the folder "/opt/config/ssl".
+
+Then, when configuring the debezium connector, we need to also mount these certificates and properly configure the connector:
+
+```yaml
+  connect:
+    image: quay.io/debezium/connect:${DEBEZIUM_VERSION}
+    ports:
+      - "8083:8083"
+    environment:
```
jpechane, line 40: IMHO you also need producer and consumer prefixes, see https://docs.confluent.io/platform/current/security/protect-data/encrypt-tls.html#encryption-ssl-connect

kafka-ssl/README.md
Sgitario: Because the certificate I use validates that the host is localhost. We can configure Kafka to disable the hostname validation as well.

Sgitario:
@jpechane I tried the changes you suggest in the comments: (1) provide the PRODUCER_ and also the CONSUMER_ properties and (2) use the MySQL connector, so the schema history is used, but still I'm getting authenticated errors:

```
[kafka]     | 2024-12-12 14:41:52,431 - INFO  [data-plane-kafka-network-thread-1-ListenerName(SSL)-SSL-5:Selector@636] - [SocketServer listenerType=ZK_BROKER, nodeId=1] Failed authentication with /10.89.5.16 (channelId=10.89.5.16:9092-10.89.5.16:59652-44) (SSL handshake failed)
```

So, there must still be some misconfiguration to get Debezium connector work with the Kafka using SSL.

kafka-ssl/README.md
jpechane: Intersting, I am getting `Error response from daemon: conflicting options: host type networking can't be used with links. This would result in undefined behavior` when I try it with the option, but maybe we can get rid of it and have a certificate generated for hostname `kafka`?

jpechane: In that case I believe your networking and advertisment setup will simplify a lot too.

Sgitario:
@jpechane I found why the example was not working, two reasons:
- I was missing the "ssl.key.password" property
- I was not correctly configuring the producer and consumer properties (I was using "PRODUCER_XXX", when it should be "CONNECT_PRODUCER_XXX" instead). 

I see that the Debezium connector image is what is binding the above properties to the Kafka connector properties in https://github.com/debezium/container-images/blob/5e7256ac849b9a1b233135bee101bee14ac5f194/connect-base/3.0/docker-entrypoint.sh#L311

This should be documented in the Debezium site, which it's also part of this JIRA task scope.

About the comment to use an MySQL connector, what about to add it as part of another JIRA ticket plus covering these two examples using an GitHub action workflow as well?

jpechane:
@Sgitario Applied, thanks!
Feel free to follow-up with additional Jiras.
WRT the documentation - it is in fact there https://github.com/debezium/container-images/tree/main/connect-base/3.0#others - ut yes it is hidden and needs to be more visible.

PR #376, opened 2025-04-10
[debezium-rag] Quarkus RAG with Milvus store

jpechane:
## Checklist


jpechane:
@vjuranek Please take a look. I have it running locally. When the AI related PRs I created are merged I'll publish distinct tag based on 3.2 nightly and will use it in demo.
I'll work on README side-by-side with blogpost. Please take a look at current readme. it describes the demo steps. Please let me know it it makes sense to you.

jpechane:
@vjuranek The server was pushed to quay. The demo is now testable.

vjuranek:
Yes, command and readme makes sense. Really cool demo @jpechane !

Just to be clear: I haven't been able to make it working locally yet, as Milvus seems to fall into some crash-recovery loop and doesn't seem to start and I'm not able to create even initial record there. I'll dig into it, but it doesn't see to be related to anything in the demo itself, probably some issue in my env or something similar.

PR #379, opened 2025-04-22
DBZ-8945: Apicurio registry configuration should include instructions for confluent compatibility mode

kmos:
### Context
The documentation on Avro schema registry has two sections - Apicurio registry and Confluent registry.   However it doesn't have anything for Apicurio in "compatibility mode". 

### Changes

we added a new section and configuration that enable the compatibility mode.


jpechane:
@kmos Applied, thanks

PR #389, opened 2025-10-07
CQRS design pattern example

vjuranek:


cqrs/poll-results-service/src/main/java/io/debezium/examples/cqrs/ResultsResource.java
```diff
@@ -0,0 +1,50 @@
...
+    private EntityManager em;
+
+    @GET
+    @Path("{id}")
+    public List<OptionVotesEntity> getResult(Long id) {
+        try {
+            Query q = em.createNativeQuery(resultQuery, OptionVotesEntity.class);
+            List<OptionVotesEntity> res = q.getResultList();
+            for (OptionVotesEntity o : res) {
+                LOGGER.info("result: " + o.option + ", votes: " + o.votes);
```
jpechane: Please use message templating as you do in other cases instead of string concatenation.

cqrs/poll-service/src/main/java/io/debezium/examples/cqrs/entity/VoteEntity.java
```diff
@@ -0,0 +1,23 @@
...
+ */
+@Entity
+public class VoteEntity extends PanacheEntity {
+    public Long pollId;
+    public Long votedOption;
+    public Timestamp votedOn;
+
+    @Override
+    public String toString() {
+        return String.format("Vote for '{}', voted on {}", votedOption, votedOn);
```
jpechane: I don't think this is correct, you should use `%s` not `{}`

cqrs/postgres-source.json
```diff
@@ -0,0 +1,14 @@
+{
+    "name": "postgres-source",
+    "config": {
+        "connector.class": "io.debezium.connector.postgresql.PostgresConnector",
+        "tasks.max": "1",
+        "database.hostname": "postgres",
+        "database.port": "5432",
+	"database.user": "postgres",
```
jpechane: Indentation here and in other places

cqrs/postgres-source-unwrap.json
```diff
@@ -0,0 +1,17 @@
...
+        "tasks.max": "1",
+        "database.hostname": "postgres",
+        "database.port": "5432",
+	"database.user": "postgres",
+        "database.password": "postgres",
+        "database.dbname" : "postgres",
+        "topic.prefix": "polls",
+        "table.include.list": "public.optionentity, public.pollentity, public.pollentity_optionentity, public.pollentity_voteentity, public.voteentity",
+	"transforms": "unwrap",
+	"transforms.unwrap.type": "io.debezium.transforms.ExtractNewRecordState",
```
jpechane: Is not it better to apply the SMT on the sink side so other consumers can benefit from the full envelope?

jpechane (commented):
@vjuranek Thanks a lot, looks very nice!
I left couple of proposals but nothing major.

jpechane (approved):

PR #403, opened 2026-04-03
Update to DEBEZIUM_VERSION=3.4

SaidiBTW:
- Update to debezium 3.4 as requested [here](https://github.com/debezium/debezium-examples/pull/402#issuecomment-4116090585)
- Updated imports to match new Quarkus version
- Updated packages in pom.xml to new versions of docker-compose services

#402 - Changes were requested from the parent PR

## Checklist

### DIFF
--- a/auditlog/.gitignore
+++ b/auditlog/.gitignore
@@ -1,0 +1,3 @@
+target
+.quarkus
+connector-config
--- a/auditlog/README.md
+++ b/auditlog/README.md
@@ -5,37 +5,37 @@ It accompanies the blog post [Building Audit Logs with Change Data Capture and S
 
 There are two applications (based on [Quarkus](https://quarkus.io/)):
 
-* _vegetables-service_: a simple REST service for inserting and updating vegetable data into a Postgres database;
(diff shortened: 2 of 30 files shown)


SaidiBTW:
After updating the versions. The admin service has not been working according to blog instructions. It may require some refactoring.

jpechane:
@vsantonastaso Please take a look, I suppose this PR is no longer needed.

SaidiBTW:
There were big differences between the Client Code from version 1.8 to 3.4. The entire example may require a rewrite. How should we go about this?

vsantonastaso:
> @vsantonastaso Please take a look, I suppose this PR is no longer needed.

The Debezium upgrade is no longer required after this https://github.com/debezium/debezium-examples/pull/404. However,here @SaidiBTW also updated the Quarkus version and several other dependencies. Therefore, we should proceed with a refactoring to ensure the auditlog example works correctly with all the updated dependencies.

jpechane:
@SaidiBTW This should be no longer needed. Thank you for the effort. A CI action is now in place to peridically update exmplaes to use the latest final release.

PR #415, opened 2026-06-01
debezium/dbz#1809 Standardize Examples - Cache Invalidation

ilyasahsan123:
Relates to debezium/dbz#1809

## Checklist

### DIFF
--- a/.github/workflows/outbox-workflow.yml
+++ b/.github/workflows/outbox-workflow.yml
@@ -21,8 +21,9 @@ jobs:
       - name: Set up Java
         uses: actions/setup-java@v4
         with:
-          java-version: '21'
+          java-version: '17'
           distribution: 'temurin'
+
 
       - name: Cache local Maven repository
         uses: actions/cache@v4
--- a/unwrap-smt/debezium-jdbc-es/Dockerfile
+++ b/unwrap-smt/debezium-jdbc-es/Dockerfile
@@ -19,3 +19,5 @@ RUN curl -sSL -o $KAFKA_CONNECT_PLUGINS_DIR/es-connector.zip "https://api.hub.co
     unzip -d $KAFKA_CONNECT_PLUGINS_DIR $KAFKA_CONNECT_PLUGINS_DIR/es-connector.zip && \
     rm -f $KAFKA_CONNECT_PLUGINS_DIR/es-connector.zip
 
+USER kafka
+


cache-invalidation/pom.xml
```diff
@@ -10,47 +10,25 @@
...
 
     <dependencyManagement>
         <dependencies>
             <dependency>
-                <groupId>${quarkus.platform.group-id}</groupId>
-                <artifactId>${quarkus.platform.artifact-id}</artifactId>
-                <version>${quarkus.platform.version}</version>
+                <groupId>io.quarkus</groupId>
+                <artifactId>quarkus-bom</artifactId>
+                <version>${version.quarkus}</version>
```
vsantonastaso: I think that here we could avoid specifying the version since it is already defined in the parent pom. WDYT?

cache-invalidation/pom.xml
```diff
@@ -112,10 +84,15 @@
...
             <plugin>
-                <groupId>${quarkus.platform.group-id}</groupId>
+                <groupId>io.quarkus</groupId>
                 <artifactId>quarkus-maven-plugin</artifactId>
-                <version>${quarkus.platform.version}</version>
+                <version>${version.quarkus}</version>
                 <extensions>true</extensions>
+                <configuration>
+                    <systemProperties>
+                        <net.bytebuddy.experimental>true</net.bytebuddy.experimental>
```
vsantonastaso: Here it would be helpful to have a comment about why this property necessary. WDYT?

cache-invalidation/pom.xml
```diff
@@ -10,47 +10,25 @@
...
-
     <properties>
-        <compiler-plugin.version>3.13.0</compiler-plugin.version>
-        <maven.compiler.release>21</maven.compiler.release>
-        <quarkus.platform.artifact-id>quarkus-bom</quarkus.platform.artifact-id>
-        <quarkus.platform.group-id>io.quarkus.platform</quarkus.platform.group-id>
-        <quarkus.platform.version>3.15.1</quarkus.platform.version>
         <skipITs>true</skipITs>
-        <surefire-plugin.version>3.3.1</surefire-plugin.version>
+        <surefire-plugin.version>3.5.2</surefire-plugin.version>
```
Naros: For plugins, could we centralize these in the parent POM, like we do for the others?

pom.xml
```diff
@@ -32,6 +32,7 @@
         <maven.compiler.source>21</maven.compiler.source>
         <maven.compiler.target>21</maven.compiler.target>
         <maven.compiler.release>21</maven.compiler.release>
+        <maven.compiler.plugin>3.15.0</maven.compiler.plugin>
```
Naros: This defines a plugin version, ideally should be named `version.maven.compiler` or similar to align with how we define version for dependencies.

Naros (commented):
Just a few inline comments.

cache-invalidation/pom.xml
ilyasahsan123: We can’t remove it because it would cause conflicts with the JUnit version 6 and 5 versions.
Therefore, I keep specifying the Quarkus version to ensure that the JUnit version is pinned to 5.

ilyasahsan123:
hi @Naros @vsantonastaso, I've addressed your concern. Could you please take another look? Thanks

ilyasahsan123:
hi @Naros @vsantonastaso when you have a moment, could you please take another look? thanks

vsantonastaso:
> hi @Naros @vsantonastaso when you have a moment, could you please take another look? thanks

Hi @ilyasahsan123  maybe there are still some comments to handle, right?

PR #406, opened 2026-04-21
debezium/dbz#1810 Add automated E2E tests for the auditlog example.

KMohnishM:
Fixes debezium/dbz#1810

## Description

This PR introduces automated end-to-end testing for the `auditlog` example and extends the existing test runner to support more general use cases.

This is the first step toward standardizing automated testing across `debezium-examples`, based on the approach used in `postgres-failover-slots`.

Fixes #1810

#### 1. Auditlog E2E Test
- Added `auditlog/test.yaml` defining the full test workflow:
  - Start services via Docker Compose
  - Register connector
  - Trigger API request with authentication
  - Verify enriched Kafka output

#### 2. Test Runner Improvements (`scripts/run-example-test.py`)
- Refactored HTTP handling into reusable helpers
- Added support for:
  - `http_post` steps
  - Custom `headers` (for JWT-based flows)
  - `expected_status` in `http_wait`

These changes make the runner reusable across more complex examples.

#### 3. CI Integration
- updated workflow to execute the auditlog test using the Python runner

### Verification

- Tested locally (WSL environment)
- Confirmed that:
  - Services start correctly
  - Connector is registered successfully
  - Enriched events are produced as expected

## Checklist

- [X] If the changes include a new example, I added it to the list of examples in the [README.md](https://github.com/debezium/debezium-examples/blob/main/README.md) file

PR #416, opened 2026-06-02
debezium/dbz#1810 Add automated E2E tests for the cloudevents example

KMohnishM:
## Description

Fixes debezium/dbz#1810

This PR introduces automated end-to-end testing for the `cloudevents` example, utilizing the shared YAML-driven test runner established in #406.

### Changes
* **Debezium 3.5 API Updates**: Updated connector configurations (`register-postgres-json-json.json`, `register-postgres-json-avro.json`, `register-postgres-avro-avro.json`) to replace the deprecated/removed `database.server.name` with `topic.prefix`.
* **Quarkus 3.x / Java 21 Support**: Rewrote `avro-data-extractor/Dockerfile` to run on an Eclipse Temurin Java 21 base image and use the modern Quarkus `target/quarkus-app/` layout.
* **CloudEvents E2E Test Manifest**: Added `cloudevents/test.yaml` which defines the full test workflow:
  * Builds and starts all services via Docker Compose (PostgreSQL, Kafka, Connect, Schema Registry, Schema Registry UI, and `avro-extractor` Streams application).
  * Registers the `inventory-connector-json-json` connector configured with the CloudEvents Converter.
  * Modifies a customer's first name in the PostgreSQL source database.
  * Consumes the `dbserver1.inventory.customers` Kafka topic to verify that the CDC event is wrapped in a CloudEvents envelope (asserting on `specversion`) and contains the correct update.
* **CI Workflow**: Updated `.github/workflows/cloudevents.yml` to trigger on changes to the example or test runner, set up Python 3.11, install dependencies, and run the E2E test.

### Verification
* Verified compiling the `avro-data-extractor` Quarkus application.
* Tested and passed the E2E test execution locally using `python3 scripts/run-example-test.py cloudevents`.

## Checklist

### DIFF
--- a/cloudevents/avro-data-extractor/Dockerfile
+++ b/cloudevents/avro-data-extractor/Dockerfile
@@ -1,5 +1,13 @@
-FROM fabric8/java-jboss-openjdk8-jdk
-ENV JAVA_OPTIONS='-Dquarkus.http.host=0.0.0.0 -Dquarkus.http.port=${PORT}'
-COPY target/lib/* /deployments/lib/
-COPY target/*-runner.jar /deployments/app.jar
-ENTRYPOINT [ "/deployments/run-java.sh" ]
+FROM eclipse-temurin:21-jre
+
+ENV PORT=8079
+ENV JAVA_OPTIONS="-Dquarkus.http.host=0.0.0.0"
+
+COPY target/quarkus-app/lib/ /deployments/lib/
+COPY target/quarkus-app/*.jar /deployments/
+COPY target/quarkus-app/app/ /deployments/app/
+COPY target/quarkus-app/quarkus/ /deployments/quarkus/
+
+EXPOSE 8079
+
+ENTRYPOINT [ "sh", "-c", "java $JAVA_OPTIONS -Dquarkus.http.port=${PORT} -jar /deployments/quarkus-run.jar" ]
--- a/cloudevents/test.yaml
+++ b/cloudevents/test.yaml
@@ -1,0 +1,50 @@
+name: cloudevents
+description: Tests Debezium PostgreSQL connector with CloudEvents structured mode (JSON format)
+env_file: ../.env
+compose_file: docker-compose.yaml
+
+cleanup:
+  type: docker_compose_down
+  volumes: true
+
+steps:
+  - name: Build all images
+    type: docker_compose_build
+
+  - name: Start all services
+    type: docker_compose_up
+    detach: true
+
+  - name: Wait for Kafka Connect REST API to be ready
+    type: http_wait
+    url: http://localhost:8083/connectors
+    timeout_seconds: 120
+    interval_seconds: 5
+
+  - name: Register Postgres connector
+    type: http_put
+    url: http://localhost:8083/connectors/inventory-connector-json-json/config
+    body_file: register-postgres-json-json.json
+    expected_status: [200, 201]
+
+  - name: Wait for connector to be RUNNING
+    type: http_wait
+    url: http://localhost:8083/connectors/inventory-connector-json-json/status
+    timeout_seconds: 60
+    interval_seconds: 5
+    expected_json_path: connector.state
+    expected_value: RUNNING
+
+  - name: Modify records in Postgres database
+    type: docker_compose_exec
+    service: postgres
+    command: 'psql postgresql://postgres:postgres@localhost:5432/postgres -c "UPDATE inventory.customers SET first_name = ''Sarah'' WHERE id = 1001;"'
+
+  - name: Verify CloudEvent envelope and content in Kafka
+    type: kafka_consume
+    topic: dbserver1.inventory.customers
+    timeout_seconds: 60
+    expected_content:
+      - "specversion"
+      - "Sarah"
+      - '"iodebeziumop":"u"'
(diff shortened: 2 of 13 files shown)


cloudevents/avro-data-extractor/Dockerfile
```diff
@@ -1,5 +1,12 @@
-FROM fabric8/java-jboss-openjdk8-jdk
-ENV JAVA_OPTIONS='-Dquarkus.http.host=0.0.0.0 -Dquarkus.http.port=${PORT}'
-COPY target/lib/* /deployments/lib/
-COPY target/*-runner.jar /deployments/app.jar
-ENTRYPOINT [ "/deployments/run-java.sh" ]
+FROM eclipse-temurin:21-jre
+
+ENV JAVA_OPTIONS="-Dquarkus.http.host=0.0.0.0 -Dquarkus.http.port=8079"
```
Naros: Could we maintain port configurability?

cloudevents/test.yaml
```diff
@@ -0,0 +1,53 @@
...
+    type: kafka_consume
+    topic: dbserver1.inventory.customers
+    timeout_seconds: 60
+    expected_content: "specversion"
+
+  - name: Verify change content in CloudEvent in Kafka
+    type: kafka_consume
+    topic: dbserver1.inventory.customers
+    timeout_seconds: 30
+    expected_content: "Sarah"
```
Naros: If the intent here is to test that the update is a CloudEvents event and has Sarah, then I believe we should consider allowing a way to define a single Kafka read with multiple test assertions.  WDYT?

KMohnishM: Agreed @Naros , that's a much cleaner approach. I've updated the Python test runner (`scripts/run-example-test.py`) to support a list of expected strings under `expected_content` so we can run multiple assertions in a single topic read. I've consolidated the `cloudevents/test.yaml` steps to do this.

KMohnishM:
@Naros 
I have updated the PR to resolve the CI pipeline failure and address feedback for the E2E test setup. 

To keep the history clean, I have structured this into two separate commits:

1. **Commit 1: E2E Test Automation (`cloudevents` Example)**
   - Updated the docker compose setup and Dockerfile for `avro-data-extractor` to preserve port configurability using a dynamic `PORT` environment variable (`-Dquarkus.http.port=${PORT}`).
   - Refactored `run-example-test.py` to allow consuming and verifying multiple expected strings (`expected_content`) within a single topic-reading cycle.
   - Combined the verification steps in `test.yaml` to run assertion on both `specversion` and `Sarah` in one check step.

2. **Commit 2: Outbox Java 21 Compatibility Fix**
   - **Byte Buddy Upgrade (`outbox/pom.xml`)**: Overrode transitive `byte-buddy` and `byte-buddy-agent` versions to `1.14.18` under `<dependencyManagement>`. This prevents the build crash when Quarkus performs bytecode enhancement on Java 21 (which was caused by upstream's upgrade to Java 21 compilation targets).
   - **Runtime Container Upgrade (`order-service` & `shipment-service` Dockerfiles)**: Upgraded the JRE runtime in the Docker images from Java 11 to Java 21 (`java-21-openjdk-headless`). This resolves the `UnsupportedClassVersionError` that occurred when loading Debezium 3.5.0.Final dependencies (which target Java 17/21 bytecode) under Java 11.

outbox/pom.xml
```diff
@@ -26,6 +27,16 @@
 
   <dependencyManagement>
     <dependencies>
+      <dependency>
+        <groupId>net.bytebuddy</groupId>
+        <artifactId>byte-buddy</artifactId>
+        <version>1.14.18</version>
```
Naros: Should we define these centrally in the parent pom?

KMohnishM: Yes, that makes complete sense @Naros ! I have moved the byte-buddy and byte-buddy-agent dependency overrides centrally to the dependency Management block of the root parent pom.xml, and removed the duplicate overrides from outbox/pom.xml. 

I have updated the commits and force-pushed the branch.

cloudevents/test.yaml
```diff
@@ -0,0 +1,49 @@
...
+  - name: Modify records in Postgres database
+    type: docker_compose_exec
+    service: postgres
+    command: 'psql postgresql://postgres:postgres@localhost:5432/postgres -c "UPDATE inventory.customers SET first_name = ''Sarah'' WHERE id = 1001;"'
+
+  - name: Verify CloudEvent envelope and content in Kafka
+    type: kafka_consume
+    topic: dbserver1.inventory.customers
+    timeout_seconds: 60
+    expected_content:
```
vsantonastaso, line 47: Is it possible to check the envelope more thoroughly here? 
For example, it would be useful to check that  "specversion" and "Sarah" are both in the same message, the operation type (update and not insert/read)

vsantonastaso (approved):

cloudevents/test.yaml
KMohnishM: Yes, that's a great suggestion! I've updated the test runner to assert that all expected substrings (e.g., specversion, Sarah, and the operation type) exist within the *same* message (on a single line) rather than scattered anywhere in the consumer output. I also added a check for `"iodebeziumop":"u"` to verify that it is indeed a structured CloudEvents update operation.

I have updated the commits and force-pushed the updates.

Naros (approved):

Naros:
Applied, thanks @KMohnishM.

PR #424, opened 2026-07-09
debezium/dbz#1810 Add E2E test for engine-wasm example

KMohnishM:
### Description
This PR adds automated End-to-End (E2E) testing for the `engine-wasm` example, ensuring it functions correctly after Debezium upgrades.

**Key Changes:**
* **E2E Test Definition:** Created `engine-wasm/test.yaml` containing the complete test pipeline including database startup, WASM engine launch, data update triggers, and log verification.
* **Chicory WASM Print Support:** Updated `ChangeDataSender.java` to define and register the `env.putchar` host function in the Chicory runtime. This allows TinyGo compiled WASM modules to write to standard output.
* **Privilege Fix:** Corrected the MySQL update command in `test.yaml` to use write-eligible credentials (`mysqluser` / `mysqlpw`) instead of the read-only replication credentials (`debezium` / `dbz`).
* **Runner Shell Protection:** Improved the `pkill` command patterns (using bracket regex `ChangeDataSender` / `exec-maven-plugin`) to prevent the script from killing its own execution shell process.

### DIFF
--- a/engine-wasm/docker-compose.yaml
+++ b/engine-wasm/docker-compose.yaml
@@ -1,0 +1,10 @@
+version: '3.8'
+services:
+  mysql:
+    image: quay.io/debezium/example-mysql:${DEBEZIUM_VERSION}
+    ports:
+      - "3306:3306"
+    environment:
+      - MYSQL_ROOT_PASSWORD=debezium
+      - MYSQL_USER=mysqluser
+      - MYSQL_PASSWORD=mysqlpw
--- a/engine-wasm/docker-compose.yaml
+++ b/engine-wasm/docker-compose.yaml
@@ -1,4 +1,3 @@
-version: '3.8'
 services:
   mysql:
     image: quay.io/debezium/example-mysql:${DEBEZIUM_VERSION}
--- a/engine-wasm/src/main/java/io/debezium/examples/wasm/ChangeDataSender.java
+++ b/engine-wasm/src/main/java/io/debezium/examples/wasm/ChangeDataSender.java
@@ -7,7 +7,10 @@ import java.util.concurrent.Executors;
 import com.dylibso.chicory.runtime.ImportMemory;
 import com.dylibso.chicory.runtime.ImportValue;
 import com.dylibso.chicory.runtime.ImportValues;
+import com.dylibso.chicory.runtime.HostFunction;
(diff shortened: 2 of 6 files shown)


engine-wasm/src/main/java/io/debezium/examples/wasm/ChangeDataSender.java
```diff
@@ -96,6 +111,16 @@ public void run() {
     private void sendRecord(ChangeEvent<String, String> record) {
         LOGGER.debug("Passing change event key = '{}', value = '{}' to WASM module", record.key(), record.value());
 
+        // Log the message reception matching cdc.go format to satisfy E2E test verification
+        String id = "";
+        java.util.regex.Pattern p = java.util.regex.Pattern.compile("\"id\"\\s*:\\s*(\\d+)");
```
Naros: Could we move the Pattern compile to a static member variable?

engine-wasm/docker-compose.yaml
```diff
@@ -0,0 +1,10 @@
+version: '3.8'
```
Naros: We can drop `version` here, it's not needed in v2 of the YAML format.

Naros (commented):
Hi @KMohnishM LGTM, just a few inline points and its good to go.

KMohnishM:
@Naros I have updated the PR according to comments , Thanks for the feedback!

engine-wasm/test.yaml
```diff
@@ -0,0 +1,55 @@
+name: engine-wasm
+description: Tests Debezium Engine streaming database changes to a WASM application (Chicory / TinyGo)
+
+env_file: ../.env
+compose_file: docker-compose.yaml
+
+cleanup:
+  type: docker_compose_down
+  volumes: true
```
Naros: If the test fails when running the steps in lines 46-48 (pkill), this can leave a rogue Java process running. On CI, this resolves itself when the runner exits, but locally it leaves orphaned processes. Would it be possible to add a pkill step before the compose teardown in cleanup, or restructure cleanup so both are handled?

Naros (changes requested):

Naros (approved):

Naros:

PR #426, opened 2026-07-11
debezium/dbz#1810 Add E2E test for debezium-python example

KMohnishM:
Fixes debezium/dbz#1810

## Description

This PR adds automated End-to-End (E2E) testing for the `debezium-python` example, which demonstrates Debezium's Connect mode (zero JSON overhead, passing raw Java `SourceRecord` objects directly to Python via JPype).

Additionally, it resolves a blocker where the postgres container readiness wait check was hanging indefinitely on newer versions of `testcontainers`.

#### **Key Changes**
1. **Container Startup Fix (`debezium-python/connect_mode_test.py`)**: Removed the custom `wait_for_logs` override patch checking for `.*PostgreSQL init process complete.*` which was hanging. Allowing `testcontainers` to use its default connection-readiness wait strategy (`ExecWaitStrategy` executing `psql` directly inside the container) resolves the hang.
2. **E2E Test Definition (`debezium-python/test.yaml`)**: Created the test pipeline definition to install dependencies, run `setup_jars.py` to download Maven dependency JARs, and run `connect_mode_test.py` to verify the execution.
3. **GitHub Action Workflow (`.github/workflows/debezium-python-workflow.yml`)**: Added the automated CI workflow to trigger on code changes in the `debezium-python` example.
## Checklist

### DIFF
--- a/.github/workflows/debezium-python-workflow.yml
+++ b/.github/workflows/debezium-python-workflow.yml
@@ -1,0 +1,44 @@
+name: Build [debezium-python]
+
+on:
+  push:
+    paths:
+      - 'debezium-python/**'
+      - '.github/workflows/debezium-python-workflow.yml'
+      - 'scripts/run-example-test.py'
+  pull_request:
+    paths:
+      - 'debezium-python/**'
+      - '.github/workflows/debezium-python-workflow.yml'
+      - 'scripts/run-example-test.py'
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Set up Java
+        uses: actions/setup-java@v4
+        with:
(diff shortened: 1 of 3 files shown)


Naros (approved):

Naros (changes requested):
It seems like the test ran for 6h and never finished. Could you take a look why?

KMohnishM:
Ah , I think it was because the engine was configured with self.max_events = 5. However, the sample database snapshot contains exactly 4 records (Sally, George, Edward, Anne). Since no manual streaming writes are triggered during the E2E test, the test runner waited indefinitely for a 5th event that never arrived, resulting in a hang and eventual cancellation by GitHub Actions.

Naros (approved):

Naros:

PR #428, opened 2026-07-17
debezium/dbz#2249 replace Bitnami images with open Soldevelo equivalents

jkondrat-sd:
Closes debezium/dbz#2249

## Fix: replace restricted Bitnami images

Since **August 28, 2025**, Bitnami distributes most container images only through paid VMware Tanzu subscriptions. Pulling them without a valid subscription may fail silently or return stale image layers, which is a supply-chain reliability concern.

This PR replaces every affected reference with the corresponding image from [SolDevelo/containers](https://github.com/SolDevelo/containers) - a community fork of `bitnami/containers` that publishes identical, freely accessible images to Docker Hub under the `soldevelo/` namespace.

No configuration changes beyond the image tag itself are required.

## Replaced references

- `bitnami/redis:7.0` → [`soldevelo/redis:7.0`](https://hub.docker.com/r/soldevelo/redis)

### DIFF
--- a/debezium-server/debezium-server-mysql-redis-pubsub/docker-compose.yml
+++ b/debezium-server/debezium-server-mysql-redis-pubsub/docker-compose.yml
@@ -9,7 +9,7 @@ services:
       - MYSQL_PASSWORD=mysqlpw
 
   redis:
-    image: bitnami/redis:7.0
+    image: soldevelo/redis:7.0
     ports:
       - 6379:6379
     environment:


Naros:
Hi @jkondrat-sd, please also raise a GH issue to track this change in https://github.com/debezium/dbz/issues, and apply the issue key as a prefix on the commit. Thanks. 

In addition, if this was created using autonomous AI, I'd recommend reading our AI Policy:
https://github.com/debezium/debezium/blob/main/AI_USAGE_POLICY.md

jkondrat-sd:
> Hi @jkondrat-sd, please also raise a GH issue to track this change in https://github.com/debezium/dbz/issues, and apply the issue key as a prefix on the commit. Thanks.
> 
> In addition, if this was created using autonomous AI, I'd recommend reading our AI Policy: https://github.com/debezium/debezium/blob/main/AI_USAGE_POLICY.md

Thanks, I've opened a tracking issue and updated the commit to include the issue number. This PR was not created using AI; it was only used for wording improvements.

debezium-server/debezium-server-mysql-redis-pubsub/docker-compose.yml
```diff
@@ -9,7 +9,7 @@ services:
       - MYSQL_PASSWORD=mysqlpw
 
   redis:
-    image: bitnami/redis:7.0
+    image: soldevelo/redis:7.0
```
vjuranek, line 12: In the rest of the codebase (e.g. Debezium server tests for Redis), we use `mirror.gcr.io/library/redis`. IMHO it would be better to use this one - we unify it across the code base and also we won't depend on another company which may discontinue it anytime in the future (without us noticing it).

PR #429, opened 2026-07-19
debezium/dbz#1810 Add E2E test for mysql-replication example

KMohnishM:
Fixes debezium/dbz#1810

## Description

Here is the updated PR description for the `mysql-replication` example:

This PR adds automated End-to-End (E2E) testing for the `mysql-replication` example, which verifies the Debezium MySQL connector's failover capability from a primary/master node to its replica database and back.

Additionally, this PR fixes a timing/startup race condition that occurs during replication initialization on Docker Compose startup and container restarts.

#### **Key Changes**

1. **E2E Test Definition (`mysql-replication/test.yaml`)**:
   * Creates the test workflow executing the complete replication failover scenario:
     1. Starts MySQL master-replica services.
     2. Deploys the MySQL connector pointing to `mysql-replica` and checks the initial snapshot ingest.
     3. Inserts customer records on the master node and verifies they are streamed to Kafka.
     4. Stops the master node, updates the connector config to point to `mysql-replica` directly.
     5. Restarts the master node, writes subsequent data, and verifies replication/streaming resume successfully.
     6. Switches the connector configuration back to `mysql_master`.

2. **Replication Race Condition Fix (`mysql-replication/docker-compose.yaml`)**:
   * Added a `healthcheck` to the `mysql_master` service utilizing `mysqladmin ping`.
   * Modified `mysql_replica` to depend on `mysql_master` under the `service_healthy` condition. This ensures that the replica database is started only after the primary database is fully booted and replication-ready, avoiding connection retries and replica sync failures.

3. **Looping Readiness Checks (`mysql-replication/test.yaml`)**:
   * Replaced single-run `mysqladmin ping` checks with robust `until` loop checks (`until mysqladmin ping -u root -pdebezium --silent; do sleep 1; done`) to handle container initialization time gracefully during startups and restarts.

4. **Runner Script Improvements (`scripts/run-example-test.py`)**:
   * Enhanced `step_kafka_consume` to correctly support list-based expected content across distinct message streams.
   * Restructured cleanup handlers to support sequentially running multiple cleanup steps.

5. **CI Automation (`.github/workflows/mysql-replication-workflow.yml`)**:
   * Setup GitHub Actions workflow to run the example test pipeline on code changes.

## Checklist

### DIFF
--- a/mysql-replication/test.yaml
+++ b/mysql-replication/test.yaml
@@ -1,0 +1,112 @@
+name: mysql-replication
+description: Tests Debezium MySQL connector with master-replica failover and switchback
+env_file: ../.env
+compose_file: docker-compose.yaml
+
+cleanup:
+  type: docker_compose_down
+  volumes: true
+
+steps:
+  - name: Ensure clean state before starting
+    type: docker_compose_down
+    volumes: true
+
+  - name: Start all services
+    type: docker_compose_up
+    detach: true
+
+  - name: Wait for Kafka Connect REST API to be ready
+    type: http_wait
+    url: http://localhost:8083/connectors
+    timeout_seconds: 120
+    interval_seconds: 5
+
+  - name: Wait for MySQL master to be ready
+    type: docker_compose_exec
+    service: mysql_master
+    command: bash -c "for i in {1..60}; do mysqladmin ping -u root -pdebezium --silent && exit 0; sleep 1; done; exit 1"
+
+  - name: Wait for MySQL replica to be ready
+    type: docker_compose_exec
+    service: mysql_replica
+    command: bash -c "for i in {1..60}; do mysqladmin ping -u root -pdebezium --silent && exit 0; sleep 1; done; exit 1"
+
+  - name: Deploy Debezium MySQL connector
+    type: http_post
+    url: http://localhost:8083/connectors
+    body_file: register-mysql.json
+    expected_status: [200, 201]
+
+  - name: Wait for connector to be RUNNING
+    type: http_wait
+    url: http://localhost:8083/connectors/inventory-connector/status
+    timeout_seconds: 60
+    interval_seconds: 5
+    expected_json_path: connector.state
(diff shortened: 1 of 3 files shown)


Naros:
Hi @KMohnishM your (4) point in the description mentions some changes, but I did not see such files in the PR. Is it possible the PR is missing some files?

KMohnishM:
Hi @Naros, the runner script improvements (scripts/run-example-test.py) mentioned in point (4) were actually already merged into the upstream main branch via a previous pull request. Since they are already present on the base branch (debezium:main), they don't show up in the file diff for this PR. Sorry for the confusion , Thank you!

mysql-replication/test.yaml
```diff
@@ -0,0 +1,112 @@
...
+  - name: Restart master server
+    type: docker_compose_up
+    detach: true
+    services:
+      - mysql_master
+
+  - name: Wait for master to be ready after restart
+    type: docker_compose_exec
+    service: mysql_master
+    command: bash -c "until mysqladmin ping -u root -pdebezium --silent; do sleep 1; done"
```
Naros: If the `mysql_master` never starts, this could exhaust a worker for 6h. We also do this elsewhere in this YAML. I wonder if we shouldn't think about capping the iteration rather than it being uncapped. WDYT?

KMohnishM: That's a great point! I've replaced the uncapped `until` loop checks with a capped `for` loop that attempts up to 60 pings (sleeping 1 second between attempts) before exiting with a failure status `1` if MySQL fails to start:
 
 ```yaml
> command: bash -c "for i in {1..60}; do mysqladmin ping -u root -pdebezium --silent && exit 0; sleep 1; done; exit 1"
 ```
 
 I have updated the branch and force-pushed the changes. Thank you!

Naros (approved):

Naros:

PR #435, opened 2026-07-30
debezium/dbz#1809 Standardize Examples — debezium-quarkus-extension-service

ilyasahsan123:
## Checklist

### DIFF
--- a/debezium-quarkus-extension-service/docker-compose.yaml
+++ b/debezium-quarkus-extension-service/docker-compose.yaml
@@ -1,6 +1,6 @@
 services:
   postgres:
-    image: quay.io/debezium/example-postgres:3.0
+    image: quay.io/debezium/example-postgres:latest
     ports:
       - 5432:5432
     environment:
--- a/debezium-quarkus-extension-service/pom.xml
+++ b/debezium-quarkus-extension-service/pom.xml
@@ -12,37 +12,11 @@
     <artifactId>debezium-quarkus-extension-service</artifactId>
     <version>1.0.0-SNAPSHOT</version>
 
-    <properties>
-        <compiler-plugin.version>3.14.0</compiler-plugin.version>
-        <maven.compiler.release>21</maven.compiler.release>
-        <quarkus.platform.artifact-id>quarkus-bom</quarkus.platform.artifact-id>
-        <quarkus.platform.group-id>io.quarkus.platform</quarkus.platform.group-id>
-        <quarkus.platform.version>3.25.0</quarkus.platform.version>
-        <skipITs>true</skipITs>
-        <surefire-plugin.version>3.5.3</surefire-plugin.version>
-    </properties>
-
-    <dependencyManagement>
-        <dependencies>
-            <dependency>
-                <groupId>${quarkus.platform.group-id}</groupId>
-                <artifactId>${quarkus.platform.artifact-id}</artifactId>
-                <version>${quarkus.platform.version}</version>
-                <type>pom</type>
-                <scope>import</scope>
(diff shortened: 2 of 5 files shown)


ilyasahsan123:
hi @vjuranek , could you please take a loot at this pull request? thanks

vjuranek (approved):
LGTM (merging shouldn't have impact, but let's wait for the release to finish before merging)

debezium-quarkus-extension-service/docker-compose.yaml
```diff
@@ -1,6 +1,6 @@
 services:
   postgres:
-    image: quay.io/debezium/example-postgres:3.0
+    image: quay.io/debezium/example-postgres:3.5
```
Naros: I wonder whether we should use `latest` instead of `3.5` to minimize maintenance.

debezium-quarkus-extension-service/pom.xml
```diff
@@ -12,37 +12,12 @@
...
+        <!-- Pinned to 3.4.3.Final because it's the latest version-->
         <dependency>
             <groupId>io.debezium.quarkus</groupId>
             <artifactId>quarkus-debezium-postgres</artifactId>
-            <version>${version.debezium}</version>
-        </dependency>
-        <dependency>
-            <groupId>io.quarkus</groupId>
-            <artifactId>quarkus-arc</artifactId>
+            <version>3.4.3.Final</version>
```
Naros: Do we need to pin to 3.4.3.Final or could we use the Debezium version from the parent POM?

Naros (commented):
Hi @ilyasahsan123 I've just spotted two small things.

debezium-quarkus-extension-service/pom.xml
ilyasahsan123: We couldn’t use the Debezium version from the parent pom for this artifact because the latest version is 3.4.3.Final. Find the details [here](https://mvnrepository.com/artifact/io.debezium.quarkus/quarkus-debezium-postgres)

Therefore, I propose pinning the artifact to version 3.4.3.Final.

debezium-quarkus-extension-service/docker-compose.yaml
ilyasahsan123: I agree with you, and I’ve addressed it.

debezium-quarkus-extension-service/pom.xml
jpechane: @ilyasahsan123 Maybe we should switch to a new artifact then? see https://mvnrepository.com/artifact/io.debezium.quarkus/debezium-quarkus-postgres

ilyasahsan123: Yes. Thanks for the hint, @jpechane. 
I’ve addressed your comments. Let’s wait for the CI pipeline to complete.

Naros (approved):
LGTM, but one thing we probably want to shore up @ilyasahsan123 and that is in the parent POM we pull in Debezium 3.5 and Quarkus 3.20. 

These should ideally be aligned so that we pull in the Quarkus version that Debezium 3.5 shipped with, which is 3.27.1. This will be important for any examples using the Debezium Quarkus Extensions for compatibility reasons. This can be done in a separate PR, but just thought I'd raise it.

ilyasahsan123:
Good point, Thanks for raising this, @Naros. 
Noted — I'll open a separate PR to bump the Quarkus version to 3.27.1

Naros:
Applied, thanks @ilyasahsan123 :tada:

PR #436, opened 2026-08-04
Add SQL Server to Oracle replication example

EhsanMo:
Adds an example replicating a table from SQL Server into Oracle through Kafka, using the Debezium SQL Server source connector and the Debezium JDBC sink connector.

Oracle already appears in several examples, but only ever as a source. As a destination it needs no LogMiner, ARCHIVELOG mode, supplemental logging or common user, so unlike the existing Oracle examples this one runs entirely on freely pullable images and is covered by an E2E test in CI.

The target table is created by the sink connector via schema.evolution=basic. That alone generates a CLOB for every text column, because a Kafka STRING carries no length, and a CLOB supports neither an ordinary index nor a unique constraint. The source connector therefore sets column.propagate.source.type so that the original type and length reach the sink, which then generates NVARCHAR2 columns and a primary key instead. The README documents that the propagated width doubles, as SQL Server reports an NVARCHAR(255) column as 510 bytes.

Also documented: the licensing difference between the Debezium and Confluent JDBC sink connectors, identifier case folding between the two vendors, the settings required to propagate deletes, and the need to wait for both databases to report readiness before seeding them, since Oracle creates its application user late in a first-time startup.

The Connect image adds only the Oracle JDBC driver, which is not bundled with the Debezium image although the JDBC sink connector itself is.

## Checklist


sqlserver-to-oracle/docker-compose.yaml
```diff
@@ -0,0 +1,53 @@
...
+    environment:
+      - ACCEPT_EULA=Y
+      - MSSQL_PID=Standard
+      - MSSQL_SA_PASSWORD=Password!
+      - MSSQL_AGENT_ENABLED=true
+    volumes:
+      - ./sqlserver-init:/init:ro
+
+  oracle:
+    image: gvenzl/oracle-free:23-slim-faststart
```
Naros: @jpechane, what do you think about us hosting at least a `quay.io/debezium/oracle-free:23.26.1` or even `latest` image that we periodically fetch from Oracle Container Registry and push there as a mirror? The license does allow for this IFF the image is unmodified.

The other option is to explicitly use `ghcr.io` rather than Docker Hub.

sqlserver-to-oracle/debezium-jdbc-oracle/Dockerfile
```diff
@@ -0,0 +1,19 @@
+ARG DEBEZIUM_VERSION
+FROM quay.io/debezium/connect:${DEBEZIUM_VERSION}
+
+ARG ORACLE_JDBC_VERSION=23.5.0.24.07
```
Naros: Could we add this to the root directory's `.env` file and add a step in the example that the user should source that environment file before running, just like the `${DEBEZIUM_VERSION}`? 

We've done this in other places as we've started moving to a common definition for example dependencies, and it would make updating the versions much easier and keep everything in a single place.

README.md
```diff
@@ -56,6 +56,7 @@ For getting started please check the [tutorial example](./tutorial).
 * [Postgres - Kafka signal](./postgres-kafka-signal): How to stream data into the signal topic
 * [Postgres - TOAST Column Values](./postgres-toast): Dealing With Postgres TOAST Column Values
 * [SQL Server - Replication](./sql-server-read-replica): How to deploy the topology of services to stream from SQL Server read-only replica
+* [SQL Server - Streaming to Oracle](./sqlserver-to-oracle): How to capture events from a **SQL Server** database and stream them into an **Oracle** database using the Debezium JDBC sink connector
```
Naros:
```suggestion
* [SQL Server - Streaming to Oracle](./sqlserver-to-oracle): How to capture events from a SQL Server database and stream them into an Oracle database using the Debezium JDBC sink connector
```
Just to be consistent with other entries.

sqlserver-to-oracle/README.md
```diff
@@ -0,0 +1,243 @@
...
+by side (8 GB allocated to Docker is comfortable).
+
+Both databases are available for `linux/amd64`, and Oracle additionally for
+`linux/arm64`. On an arm64 host the SQL Server container therefore runs
+emulated, which works but makes it noticeably slower to start.
+
+## Running the example
+
+```shell
+export DEBEZIUM_VERSION=3.5
```
Naros:
```suggestion
```
You're already asking the user to set the environment using the `../.env` file, I'm not sure this is needed.

Naros:
Great PR @EhsanMo, I left a few comments above, if you could take a look as you have time.

sqlserver-to-oracle/debezium-jdbc-oracle/Dockerfile
EhsanMo: Done — moved to the root `.env` as `ORACLE_JDBC_VERSION` and wired it through as a build arg, same pattern as `DEBEZIUM_VERSION`. See 4d103a0.

README.md
EhsanMo: Applied, thanks. See 4d103a0.

sqlserver-to-oracle/README.md
EhsanMo: Good catch, removed. See 4d103a0.

sqlserver-to-oracle/docker-compose.yaml
EhsanMo: Leaving this thread open since it's a call for @jpechane on registry/hosting policy rather than something in this PR's diff — happy to make whichever change is preferred once that's decided.

Naros: Hi @EhsanMo could you proceed with ghcr.io instead, and then I think we're good to go.

EhsanMo: Hi @Naros — as you suggested, switched to `ghcr.io/gvenzl/oracle-free:23-slim-faststart`. Also rebased onto main to clear the conflict; the E2E passes on 3.6.

Naros (approved):

Naros:
Applied, thanks @EhsanMo :tada:

PR #471, opened 2026-09-19
debezium/dbz#2626 Add Milvus to PostgreSQL example

### LINKED ISSUE #2626: Debezium Example for Milvus to PostgreSQL (Vector)
## Description
Add a `milvus-to-pgvector` example that streams a Milvus collection into
PostgreSQL + pgvector via Kafka, using the Debezium Milvus source connector and the
Debezium JDBC sink connector.

## Goals
What it should cover:
- Initial snapshot (op=r) of a Milvus collection, then streaming inserts and deletes
read from Milvus's own Kafka message queue.
- Embeddings landing in Postgres as a native pgvector column
(the sink creates the table, mapping the vector to halfvec).
- How a Milvus upsert surfaces on the topic: op=d + tombstone + op=c at one commit
timestamp (the same source.tso), delete first, leaving exactly one row in Postgres.
No op=u, for any key.
- Vector fidelity: pgvector <=> cosine distance and Milvus ANN return the same
nearest neighbours for the same query vector.
- Offset handling: pause the source, write rows, resume, and both sides converge on the
same set of primary keys.

kartikangiras:
## Description

New `milvus-to-pgvector/` example that replicates a Milvus collection into
PostgreSQL/pgvector via Kafka, using the incubating Debezium Milvus source
connector and the Debezium JDBC sink connector.

- Compose topology: Milvus (+ etcd, MinIO), Kafka, Kafka Connect, PostgreSQL with pgvector
- Source/sink connector configs, a seed script for the `products` collection, and a test.yaml
- Demonstrates snapshot + streaming of inserts/deletes, and the `FloatVector` logical type
  mapped to a native pgvector column by the sink's table auto-creation

The Milvus connector isn't released yet, so the README covers building it from
debezium/debezium-connector-milvus into `./plugins`.

fixes debezium/dbz/issues/2626

## Checklist

- [x] If the changes include a new example, I added it to the list of examples in the [README.md](https://github.com/debezium/debezium-examples/blob/main/README.md) file


Naros (commented):
Hi @kartikangiras the example LGTM, I just have a few observations we should probably shore up or file as follow-ups:

**Scope gap**: 
In the associated issue, there are 5 goals, but two are absent, and the other is likely just a documentation-related issue. 

First, the "Vector Fidelity" goal says that the example should show it working in both directions, but this only runs `pgvector` to `Milvus`, not the inverse. Second, the "Offset handling" part about pausing the source, writing the rows, resuming, and both sides converging is missing. And finally, the "Upsert" goal is shown in the README but never shown in the topic or asserted in the test.

**CI**: 
This project currently has no `.github/**` CI workflow. There should be one added. See others for an example.

**Follow-up**: 
In `test.yaml`, there is a line that checks `pg_isready`. This could hang CI indefinitely if the database never becomes ready. This isn't just an issue in this PR; other examples have it too: `ai-rag`, `debezium-server-name-mapper`, `engine-wasm`, and `monitoring`, to name a few. 

I'd recommend a follow-up to change all examples to use something like:
```
timeout 120 bash -c 'until pg_isready -U postgres; do sleep 2; done'
```
This avoids a possible indefinite hang.

milvus-to-pgvector/README.md
```diff
@@ -0,0 +1,210 @@
...
+
+Because of that build step the example is not run by a GitHub workflow like the other examples.
+`test.yaml` covers the steps below and is meant to be run locally, with
+`python3 scripts/run-example-test.py milvus-to-pgvector` from the repository root.
+
+## Running the example
+
+### Build the Milvus connector
+
+Build the connector from [debezium/debezium-connector-milvus](https://github.com/debezium/debezium-connector-milvus)
```
kmos: All this part can be removed: you can use the `SNAPSHOT` or the beta release: https://mvnrepository.com/artifact/io.debezium/debezium-connector-milvus/3.7.0.Beta2

kartikangiras: Updated.

Naros (approved):
LGTM, @kmos feel free to merge if it's good for you.

kmos (approved):

